PR #61, opened 2026-03-12
ci: add fuzz testing workflow

thieman:
## Summary

- Adds `.github/workflows/fuzz.yml`: a new workflow that runs all `Fuzz*` targets (excluding differential tests) for 30s each on every push and PR, across the `head`, `cat`, `wc`, `tail`, and `grep` builtin packages.
- Updates `.github/workflows/test.yml`: adds a fuzz seed corpus regression step to the main CI matrix so any checked-in corpus entries that cause crashes are caught on every PR, not just during fuzz runs.
- Adds `testdata/fuzz/.gitkeep` placeholders in each builtin test package to ensure corpus cache paths are stable.
- Documents fuzz corpus retention policy in `.gitignore`.

The workflow is resilient: if no `Fuzz*` functions exist yet in a package, steps skip gracefully with an informational message rather than failing. This means the workflow is immediately usable as new fuzz tests are added by other PRs.

Corpus is persisted via `actions/cache` keyed by `github.sha` with a fallback prefix key, so each run benefits from previous fuzzing sessions.

## Test plan

- [ ] Verify the fuzz workflow appears in GitHub Actions after merging.
- [ ] Confirm that packages without fuzz targets produce a graceful skip message, not a failure.
- [ ] Confirm the seed corpus step in the main CI passes on all three OS targets.
- [ ] After a fuzz run, verify corpus entries are cached and restored on subsequent runs.

---

**Original prompt:** Research formal testing approaches for rshell builtin implementations (memory benchmarking, property/fuzz testing). Implement Option 6: a GitHub Actions CI workflow for continuous fuzzing.

- Go 1.18+ `testing.F`/`f.Fuzz()` is coverage-guided and persists corpus in `testdata/fuzz/<FuncName>/`. Running with `-run '^Fuzz'` (no `-fuzz=`) executes only seed corpus entries as regular tests — ideal for CI regression checking.
- GNU coreutils uses oss-fuzz (external); uutils/coreutils uses cargo-fuzz. Neither has the integrated CI-fuzzing-per-PR pattern rshell now gets.
- GitHub Actions `actions/cache` with SHA key + prefix fallback allows corpus to grow across runs without stale-key misses.
- `|| true` was deliberately avoided: a failure accumulator (`FAILED=0`/`FAILED=1`) + `exit $FAILED` correctly surfaces fuzzer crashes while still running all targets.
- All action SHAs are pinned (not mutable tags) per supply-chain security best practice.
- The `-fuzztime=0s` flag is invalid without `-fuzz=` — seed corpus regression step uses `-run '^Fuzz'` only.

**Key design decisions:**
- Matrix: `head`, `cat`, `wc`, `tail` (grep removed — no fuzz tests yet, would be a no-op)
- 30s per target (fast enough for CI, long enough to find shallow bugs)
- `fail-fast: false` so all targets run even if one crashes
- Workflow triggers: `push` to `main` + `pull_request` only (not on schedules/forks to control cost)


.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,76 @@
+name: Fuzz Tests
+
+on:
+  push:
+    branches: ['**']
```
thieman: <sub><sub>P3 Badge</sub></sub> **Workflow triggers on all branches and all PRs — broader scope than other workflows**

`push: branches: ['**']` means the 5-job fuzz matrix runs on every branch push across the repo, including WIP/draft branches. Combined with the unfiltered `pull_request:` trigger (no `branches:` constraint, unlike `test.yml` and `compliance.yml` which filter to `[main]`), this could consume significant Actions minutes for active development branches that are not close to merge.

This is a cost/noise concern rather than a security issue.

**Remediation (optional):** Consider restricting the `pull_request` trigger to `branches: [main]` to match the other workflows, and/or adding `paths:` filters to only run when `interp/` is changed. The `push: branches: ['**']` is more defensible if you want continuous corpus building on all branches.

PR #62, opened 2026-03-12
Add differential fuzz tests comparing rshell builtins against GNU coreutils

thieman:
## Summary

- Adds differential fuzz tests for `head`, `cat`, `wc`, and `tail` that compare rshell builtin output byte-for-byte against GNU coreutils
- Tests are gated behind `RSHELL_BASH_TEST=1` and skipped (not failed) when the GNU tool is unavailable (`exec.LookPath`)
- Seed corpus entries run as normal tests without the env var set, verifying they skip cleanly

## Tests added

| File | Fuzz functions |
|------|---------------|
| `interp/builtins/tests/head/head_differential_fuzz_test.go` | `FuzzHeadDifferentialLines` (`-n`), `FuzzHeadDifferentialBytes` (`-c`) |
| `interp/builtins/tests/cat/cat_differential_fuzz_test.go` | `FuzzCatDifferential` |
| `interp/builtins/tests/wc/wc_differential_fuzz_test.go` | `FuzzWcDifferentialLines` (`-l`), `FuzzWcDifferentialWords` (`-w`), `FuzzWcDifferentialBytes` (`-c`) |
| `interp/builtins/tests/tail/tail_differential_fuzz_test.go` | `FuzzTailDifferential` (`-n`) |

## Design notes

- GNU tools run with `cmd.Dir` set to the same temp directory as rshell so relative filenames in command output match exactly (important for `wc` which includes the filename)
- On macOS with Homebrew coreutils, the `g`-prefixed tools are used (e.g. `ghead`, `gwc`)
- Inputs are capped at 64KB and n values at 10000 for fuzz speed
- Sandbox errors (permission denied, not allowed) cause the comparison to be skipped rather than failed
- `//go:build !windows` since GNU tools are not available on Windows

## Test plan

- [ ] `go test ./interp/builtins/tests/... -run ^$` — verify compilation
- [ ] `go test ./interp/builtins/tests/... -run 'FuzzHeadDifferential|FuzzCatDifferential|FuzzWcDifferential|FuzzTailDifferential'` — verify seeds skip without env var
- [ ] `RSHELL_BASH_TEST=1 go test ./interp/builtins/tests/... -run 'FuzzHeadDifferential|FuzzCatDifferential|FuzzWcDifferential|FuzzTailDifferential'` — verify all seeds pass against GNU tools
- [ ] Optional: `RSHELL_BASH_TEST=1 go test -fuzz=FuzzHeadDifferentialLines -fuzztime=30s ./interp/builtins/tests/head/`

---

**Original prompt:** Research formal testing approaches for rshell builtin implementations. Implement Option 4: differential/oracle fuzz testing against GNU coreutils.

- Differential fuzzing runs the *same* input through two implementations and compares outputs — a classical oracle pattern. GNU coreutils is the reference oracle for this shell.
- The existing `RSHELL_BASH_TEST=1` gate (used by Docker bash comparison tests) was reused to avoid adding a new env var; differential fuzz tests require the GNU tool to be present locally.
- `exec.LookPath` gates each test: if the GNU tool is missing, tests skip rather than fail. This means the tests compile and run cleanly on all CI platforms (Windows included, via build tag exclusion).
- macOS ships BSD tools; Homebrew coreutils provides GNU tools prefixed with `g` (e.g. `ghead`, `gwc`, `gtail`). The `gnuCmd()` helper handles this automatically.
- `cmd.Dir` must be set to the same temp directory as rshell's sandbox root so that `wc file.txt` produces matching filenames in output from both implementations.
- Sandbox permission errors from rshell (path not under allowed root) are treated as skip rather than failure — the fuzz corpus contains inputs that may reference paths outside the temp dir.

**Key design decisions:**
- 64KB input cap (differential tests are slower due to subprocess overhead)
- n clamped to [0, 10000] to avoid slow GNU tool invocations on huge counts
- `helpers_test.go` in each package provides `cmdRunCtx` — the shared test infrastructure required by differential tests alongside unit tests
- Exits 0 and 1 are both valid for most tools; only unexpected exit codes (e.g. 2 for usage errors) trigger test failure

**Merge note:** This PR and PR #63 (native fuzz) both write to `interp/builtins/tests/<cmd>/`. Merge one first to avoid conflicts on the helper files.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,10 +43,12 @@ jobs:
         env:
           RSHELL_BASH_TEST: "1"
         run: |
+          OVERALL_STATUS=0
           for PKG in ./interp/builtins/tests/cat/ ./interp/builtins/tests/head/ ./interp/builtins/tests/tail/ ./interp/builtins/tests/wc/; do
             FUZZ_FUNCS=$(grep -r '^func Fuzz.*Differential' $PKG 2>/dev/null | sed 's/.*func \(Fuzz[^(]*\).*/\1/' | sort -u)
             for FUNC in $FUZZ_FUNCS; do
               echo "Fuzzing $FUNC in $PKG..."
-              go test -fuzz="^${FUNC}$" -fuzztime=30s $PKG -timeout 300s
+              go test -fuzz="^${FUNC}$" -fuzztime=30s $PKG -timeout 300s || OVERALL_STATUS=1
             done
           done
+          exit $OVERALL_STATUS
--- a/interp/builtins/tests/wc/testdata/fuzz/FuzzWcDifferentialWords/1c6e2e9cd7371f3e
+++ b/interp/builtins/tests/wc/testdata/fuzz/FuzzWcDifferentialWords/1c6e2e9cd7371f3e
@@ -1,0 +1,2 @@
+go test fuzz v1
+[]byte("\U00089249")
--- a/interp/builtins/wc/wc.go
+++ b/interp/builtins/wc/wc.go
@@ -289,6 +289,14 @@ func countReader(ctx context.Context, r io.Reader) (counts, error) {
 				} else if unicode.IsControl(r) {
 					// Non-whitespace control chars (C0, DEL, C1) are transparent:
 					// they do not start or end words, matching GNU wc in POSIX locale.
+				} else if unicode.Is(unicode.Zs, r) {
+					// Unicode space separators (NBSP, thin space, etc.) end words,
+					// matching GNU wc behaviour under C.UTF-8 locale.
+					lineLen++
+					inWord = false
+				} else if !unicode.IsGraphic(r) && !unicode.Is(unicode.Cf, r) && !unicode.Is(unicode.Co, r) {
+					// Cn (unassigned codepoints): transparent like control chars --
+					// they do not start or end words, matching GNU wc under C.UTF-8.
 				} else {
 					if !inWord {
 						c.words++
--- a/tests/allowed_symbols_test.go
+++ b/tests/allowed_symbols_test.go
@@ -126,10 +126,16 @@ var builtinAllowedSymbols = []string{
 	"unicode.Cc",
 	// unicode.Cf — format character category range table; pure data, no I/O.
 	"unicode.Cf",
+	// unicode.Co — private-use character category range table; pure data, no I/O.
+	"unicode.Co",
 	// unicode.Is — checks if rune belongs to a range table; pure function, no I/O.
 	"unicode.Is",
 	// unicode.IsControl — reports whether rune is a control character; pure function, no I/O.
 	"unicode.IsControl",
+	// unicode.IsGraphic — reports whether rune is defined as a graphic character; pure function, no I/O.
+	"unicode.IsGraphic",
+	// unicode.Zs — Unicode space separator category range table; pure data, no I/O.
+	"unicode.Zs",
 	// unicode.Me — enclosing mark category range table; pure data, no I/O.
 	"unicode.Me",
 	// unicode.Mn — nonspacing mark category range table; pure data, no I/O.


.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Fuzz (${{ matrix.name }})
+        run: |
+          FUZZ_FUNCS=$(grep -r '^func Fuzz' ${{ matrix.pkg }} 2>/dev/null | grep -v 'Differential' | sed 's/.*func \(Fuzz[^(]*\).*/\1/' | sort -u)
+          if [ -z "$FUZZ_FUNCS" ]; then
+            echo "No fuzz targets found in ${{ matrix.pkg }}, skipping"
+            exit 0
+          fi
+          for FUNC in $FUZZ_FUNCS; do
+            echo "Fuzzing $FUNC..."
+            go test -fuzz="^${FUNC}$" -fuzztime=30s ${{ matrix.pkg }} -timeout 120s || true
```
thieman: <sub><sub>P2 Badge</sub></sub> **`|| true` silently swallows fuzzer-discovered crashes**

Go's fuzz engine exits non-zero when it finds a crasher or a failing assertion. The `|| true` here means a real bug discovered during the 30-second fuzz run causes CI to pass silently — defeating the purpose of running fuzz tests in CI.

**Fix:** Remove `|| true`. Handle pre-existing failing corpus entries by replaying the seed corpus deterministically in the earlier `Run fuzz seed corpus` step instead.

interp/builtins/tests/cat/cat_differential_fuzz_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+		dir := t.TempDir()
+		if err := os.WriteFile(filepath.Join(dir, "input.txt"), input, 0644); err != nil {
+			t.Fatal(err)
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		rshellOut, rshellErr, rshellCode := cmdRunCtx(ctx, t, "cat input.txt", dir)
```
thieman: <sub><sub>P0 Badge</sub></sub> **`cmdRunCtx` undefined — package fails to compile**

The function `cmdRunCtx` is called here but is never defined in the `cat_test` package. All four new differential fuzz packages (`cat`, `head`, `tail`, `wc` under `interp/builtins/tests/`) have this same problem — they call `cmdRunCtx` without defining it. This causes a build failure that directly breaks the CI `Run fuzz seed corpus (regression test)` step in `test.yml`.

Verified locally:
```
$ go test -run '^$' ./interp/builtins/tests/cat/
# ...tests/cat_test
cat_differential_fuzz_test.go:93:39: undefined: cmdRunCtx
FAIL
```

**Fix:** Add a `helpers_test.go` file to each affected package defining `cmdRunCtx`. See `interp/builtins/tests/cut/cut_test.go` for the pattern (`runScriptCtx`). The helper must accept a `context.Context`, run the script in rshell with `AllowedPaths([]string{dir})`, and return stdout, stderr, exitCode.

interp/builtins/tests/tail/tail_differential_fuzz_test.go
```diff
@@ -0,0 +1,123 @@
...
+		if err := os.WriteFile(filepath.Join(dir, "input.txt"), input, 0644); err != nil {
+			t.Fatal(err)
+		}
+
+		nStr := fmt.Sprintf("%d", n)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		rshellOut, rshellErr, rshellCode := cmdRunCtx(ctx, t, fmt.Sprintf("tail -n %s input.txt", nStr), dir)
```
thieman: <sub><sub>P0 Badge</sub></sub> **`cmdRunCtx` undefined — package fails to compile**

Same as `tests/cat/`: `cmdRunCtx` is called but never defined in `tail_test`. Package fails to compile.

**Fix:** Add `helpers_test.go` with `package tail_test` defining `cmdRunCtx`.

AlexandreYang (approved):

PR #63, opened 2026-03-12
Add native Go fuzz tests for builtin commands

thieman:
## Summary

- Adds `testing.F` fuzz tests for `head`, `cat`, `wc`, `tail`, and `grep` builtins
- Each fuzz test file lives in `interp/builtins/tests/<cmd>/<cmd>_fuzz_test.go`
- Covers both file-based input and stdin-via-shell-redirection variants

## What's included

For each builtin:
- **head**: `FuzzHeadLines`, `FuzzHeadBytes`, `FuzzHeadStdin` — verifies exit codes 0/1 and that `-n K` produces at most K newlines
- **cat**: `FuzzCat`, `FuzzCatNumberLines`, `FuzzCatStdin` — verifies output exactly matches input for the plain `cat` case
- **wc**: `FuzzWc`, `FuzzWcLines`, `FuzzWcBytes`, `FuzzWcStdin` — verifies exit codes are sane
- **tail**: `FuzzTailLines`, `FuzzTailBytes`, `FuzzTailStdin` — verifies exit codes and output line/byte count invariants
- **grep**: `FuzzGrepFileContent`, `FuzzGrepStdin`, `FuzzGrepFlags` — verifies exit codes 0, 1, or 2

## Seed corpus

Each fuzz function covers edge cases:
- Empty input (`[]byte{}`)
- Input with no trailing newline
- NUL bytes (`[]byte("a\x00b\n")`)
- Buffer boundary: 4097 bytes
- All-newlines input (`[]byte("\n\n\n")`)

## Safety guards

- Inputs > 1 MB are skipped (`return`)
- Numeric params clamped: `n < 0 → return`, `n > 10000 → n = 10000`
- `context.WithTimeout(5s)` on every test to detect hangs
- Exit code assertion: must be 0 or 1 (grep also allows 2 for invalid regex)

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./interp/builtins/tests/... -run Fuzz -timeout 60s` — all seed corpus entries pass
- [ ] Optionally run full fuzzing: `go test ./interp/builtins/tests/head/... -fuzz FuzzHeadLines -fuzztime 30s`

---

**Original prompt:** Research formal testing approaches for rshell builtin implementations. Implement Option 1: native Go fuzz tests using `testing.F`.

- Go 1.18+ provides `testing.F` with `f.Add(seedValue)` for seed corpus and `f.Fuzz(func(t *testing.T, ...))` for the fuzz body. Coverage-guided, persists findings in `testdata/fuzz/<FuncName>/`.
- Running `go test -run '^FuzzXxx'` (no `-fuzz=`) executes only seed corpus entries as regular tests — zero overhead in normal CI.
- Full fuzzing: `go test -fuzz=FuzzXxx -fuzztime=30s ./pkg/` — runs in-process, no external tools needed.
- Property-based invariants are stronger than just "doesn't panic": `head -n 5` must produce ≤5 newlines; `cat` must produce output matching input byte-for-byte; grep exit code must be 0, 1, or 2.
- 1MB input cap prevents slow test runs; without it the fuzzer can generate multi-MB inputs that take seconds each.
- `context.WithTimeout(5s)` catches infinite loops / hangs that would otherwise block the fuzzer indefinitely.
- `n` clamped to 10000 prevents the fuzzer from generating `head -n 99999999` which would be valid but slow.

**Key design decisions:**
- Invariant-based assertions (not just crash detection): stronger than fuzzing for panics alone
- All fuzz tests in `interp/builtins/tests/<cmd>/` alongside existing unit tests, sharing the `cmdRun`/`cmdRunCtx` helper pattern
- `grep` fuzz tests use `FuzzGrepFlags` to explore flag combinations (e.g. `-i`, `-v`, `-c`) — important since grep has the most flag surface area

**Merge note:** This PR and PR #62 (differential fuzz) both write to `interp/builtins/tests/<cmd>/`. Merge one first to avoid conflicts on helper files.

### DIFF
--- a/interp/builtins/tests/tail/tail_fuzz_test.go
+++ b/interp/builtins/tests/tail/tail_fuzz_test.go
@@ -19,11 +19,6 @@ import (
 	"github.com/DataDog/rshell/interp/builtins/testutil"
 )
 
-func cmdRun(t *testing.T, script, dir string) (string, string, int) {
-	t.Helper()
-	return testutil.RunScript(t, script, dir, interp.AllowedPaths([]string{dir}))
-}
-
 func cmdRunCtx(ctx context.Context, t *testing.T, script, dir string) (string, string, int) {
 	t.Helper()
 	return testutil.RunScriptCtx(ctx, t, script, dir, interp.AllowedPaths([]string{dir}))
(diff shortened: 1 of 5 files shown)


interp/builtins/tests/tail/tail_fuzz_test.go
```diff
@@ -0,0 +1,156 @@
...
+func FuzzTailLines(f *testing.F) {
+	f.Add([]byte("line1\nline2\nline3\n"), int64(2))
+	f.Add([]byte{}, int64(0))
+	f.Add([]byte("no newline"), int64(1))
+	f.Add([]byte("a\x00b\nc\n"), int64(2))
+	f.Add(bytes.Repeat([]byte("x"), 4097), int64(1))
+	f.Add([]byte("\n\n\n"), int64(5))
+	f.Add(bytes.Repeat([]byte("y"), 4096), int64(1))
+	f.Add([]byte("hello\nworld\n"), int64(10))
+
```
thieman: <sub><sub>P3 Badge</sub></sub> **Missing fuzz coverage for tail `+N` offset mode**

`tail.go` implements two distinct code paths for each mode:
- Last-N mode: `readLastLines` / `readLastBytes`
- Offset mode (`+N`): `skipLines` / `skipBytes`

The current fuzz tests only exercise the last-N mode. Adding `FuzzTailLinesOffset` and `FuzzTailBytesOffset` would exercise `skipLines` and `skipBytes` with fuzz-generated input and offset values. Example:

```go
// FuzzTailLinesOffset fuzzes tail +N (skip first N-1 lines).
func FuzzTailLinesOffset(f *testing.F) {
    f.Add([]byte("line1\nline2\nline3\n"), int64(2))
    f.Fuzz(func(t *testing.T, input []byte, n int64) {
        if len(input) > 1<<20 || n < 1 || n > 10000 { return }
        dir := t.TempDir()
        // ... write input.txt ...
        ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
        defer cancel()
        _, _, code := cmdRunCtx(ctx, t, fmt.Sprintf("tail -n +%d input.txt", n), dir)
        if code != 0 && code != 1 { t.Errorf(...) }
    })
}
```

PR #109, opened 2026-03-16
feat: full GNU ls -l output with owner, group, nlink, and column alignment

matt-dz:
## Summary

Brings `ls -l` output to near-parity with GNU ls, adding columns that were previously omitted and fixing alignment.

### What changed

- **Column alignment** — right-aligns size and nlink columns using `%*s` format verb, matching GNU ls
- **Owner/group columns** — shows numeric UID/GID (not resolved names, to avoid reading `/etc/passwd` outside the sandbox)
- **Hard link count** — extracted from `Stat_t.Nlink` on Unix; on Windows, queried via `GetFileInformationByHandle` through the sandbox (`CallContext.OpenFile`). Shows `?` when unavailable
- **Total line** — prints `total <1K-blocks>` at the top of directory listings on Unix (converted from `Stat_t.Blocks` 512-byte units). Suppressed on Windows where block allocation data is unavailable (`GetFileInformationByHandleEx` requires `unsafe`, which is permanently banned)
- **Real parent stats for `..`** — `ls -la` now stats the real parent directory for `..` metadata, falling back to `.` info if the parent is outside the sandbox
- **Platform-specific files** — `owner_unix.go` and `owner_windows.go` in `builtins/ls/` with platform-appropriate implementations
- **Allowlist updates** — added `syscall.Stat_t`, `syscall.Handle`, `syscall.ByHandleFileInformation`, `syscall.GetFileInformationByHandle`, `os.O_RDONLY` to the `ls` per-command allowlist

### Security decisions

- **Numeric UID/GID only** — `os/user.LookupId` reads `/etc/passwd` or triggers NSS/LDAP outside the sandbox. Numeric IDs avoid this
- **Sandbox-routed file opens on Windows** — `getNlink` opens files through `callCtx.OpenFile` (not `os.Open`) to respect `AllowedPaths`
- **No fabricated data** — returns `?` for unknown values (nlink, owner, group) rather than plausible-but-wrong defaults like `1` or empty strings. Total line suppressed when blocks are unavailable

### Performance

- **Owner metadata cached per entry** — `fileOwner` called once per entry, results stored in `entryInfo`/`pathArg` structs. `computeColWidths` and `printEntry` read cached values (no double file opens on Windows)
- **Column width computation gated behind `-l`** — plain `ls` (without `-l`) has zero overhead from the new code

**Before:**
```
-rw-r--r-- 42 Jan  5 14:23 small.txt
-rw-r--r-- 12345678 Jan  5 14:23 big.txt
```

**After (Unix):**
```
total 96
-rw-r--r-- 1 502 20       42 Jan  5 14:23 small.txt
-rw-r--r-- 1 502 20 12345678 Jan  5 14:23 big.txt
```

**After (Windows):**
```
-rw-rw-rw- 1 ? ?       42 Jan  5 14:23 small.txt
-rw-rw-rw- 1 ? ? 12345678 Jan  5 14:23 big.txt
```

## Test plan
- [x] `go build ./...` — compiles on macOS/Linux
- [x] `GOOS=windows GOARCH=amd64 go build ./...` — cross-compiles for Windows
- [x] `go test ./...` — all tests pass
- [x] `go test ./allowedsymbols/` — allowlist verification passes
- [x] Manual: `ls -l` and `ls -lh` output compared against GNU ls via Docker (`debian:bookworm-slim`)
- [x] Manual: plain `ls` (without `-l`) does not trigger UID/GID resolution
- [x] `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash -timeout 120s` — bash comparison (long format tests skipped via `skip_assert_against_bash` due to numeric UID/GID vs resolved names)

### Test coverage added
- `TestLsLongTotalLineUses1KBlocks` — Unix: asserts total line present with correct 1K blocks
- `TestLsLongTotalLineSuppressedOnWindows` — Windows: asserts total line absent
- `TestLsLongNlinkIsValid` — all platforms: asserts nlink is `?` or >= 1, never `0`
- `column_alignment.yaml` — scenario: size column right-aligned across entries
- `total_line.yaml` — scenario: total line present (with Windows override)

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -50,6 +50,9 @@ jobs:
           - pkg: ./builtins/tests/ls/
             name: ls
             corpus_path: builtins/tests/ls
+          - pkg: ./builtins/tests/ip/
+            name: ip
+            corpus_path: builtins/tests/ip
           - pkg: ./interp/tests/
             name: interp
             corpus_path: interp/tests
--- a/SHELL_FEATURES.md
+++ b/SHELL_FEATURES.md
@@ -16,6 +16,7 @@ Blocked features are rejected before execution with exit code 2.
 - ✅ `grep [-EFGivclLnHhoqsxw] [-e PATTERN] [-m NUM] [-A NUM] [-B NUM] [-C NUM] PATTERN [FILE]...` — print lines that match patterns; uses RE2 regex engine (linear-time, no backtracking)
 - ✅ `head [-n N|-c N] [-q|-v] [FILE]...` — output the first part of files (default: first 10 lines); `-z`/`--zero-terminated` and `--follow` are rejected
 - ✅ `help` — display all available builtin commands with brief descriptions; for detailed flag info, use `<command> --help`
+- ✅ `ip [-o|-4|-6|--brief] addr|link [show] [dev IFNAME]` — show network interface addresses and link-layer info (read-only); write ops (`add`, `del`, `flush`, `set`), namespace ops (`netns`, `-n`), and batch mode (`-b`/`-B`/`--force`) are blocked
 - ✅ `sort [-rnubfds] [-k KEYDEF] [-t SEP] [-c|-C] [FILE]...` — sort lines of text files; `-o`, `--compress-program`, and `-T` are rejected (filesystem write / exec)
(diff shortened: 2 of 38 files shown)


matt-dz:
@codex review

PR #147, opened 2026-03-25
fix(ss): reduce Windows DLL buffer cap from 64 MiB to 5 MiB

thieman:
### M-3: Windows `ss` DLL buffer cap was 64 MiB

`builtins/internal/winnet/winnet_windows.go` uses a grow-loop to call `GetExtendedTcpTable` and `GetExtendedUdpTable` from `iphlpapi.dll`. The loop keeps doubling the buffer size until the DLL call succeeds, with a cap of `MaxBufSize = 64 << 20` (64 MiB).

On a system with many open sockets, this allocation could reach 64 MiB — 6× the project's 10 MiB per-invocation guideline.

## Fix

`MaxBufSize` reduced from `64 << 20` to `5 << 20`. The grow-loop behavior and bounds-checking logic are unchanged. 5 MiB is sufficient to enumerate socket tables on any realistic system; the Windows kernel itself limits the per-process socket count well within what fits in 5 MiB.

The mirror constant `MaxWinBufSize` in `builtins/ss/ss.go` (kept in sync per its comment) is updated to match.

## Test plan
- [ ] `ss` runs correctly on Windows (build compiles cleanly)
- [ ] Existing ss scenario tests pass on all platforms

### DIFF
--- a/builtins/internal/winnet/winnet_windows.go
+++ b/builtins/internal/winnet/winnet_windows.go
@@ -62,36 +62,39 @@ var tcpStateNames = map[uint32]string{
 
 // Collect enumerates all TCP and UDP sockets on Windows via iphlpapi.dll.
 // The narrow use of unsafe.Pointer is limited to two DLL call sites only.
-// If every sub-collection fails, the first error is returned so callers can
-// distinguish a real API failure from an empty socket table.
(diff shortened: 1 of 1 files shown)


thieman:
@codex review this PR

PR #149, opened 2026-03-25
fix(interp): propagate context cancellation through heredoc pipe goroutines

thieman:
### M-4: Heredoc goroutines ignore context cancellation

`interp/runner_redir.go`'s `hdocReader` function spawns goroutines to write heredoc content to pipes. These goroutines captured the heredoc content in their closure but did not accept or check the execution context (`ctx`).

If the shell context was cancelled (e.g., due to a timeout or explicit cancellation) while heredoc content was being written:
- The goroutine would continue writing to the pipe regardless
- The write would only stop when the pipe buffer filled or the reader closed the read end
- In the worst case (10 MiB heredoc, slow reader), this could keep the goroutine alive for an extended period after cancellation

This violates the principle that all execution must be interruptible via context cancellation.

## Fix

All heredoc-writing goroutines now:
1. Check `ctx.Err()` immediately on entry — if already cancelled, close the pipe and return
2. Write content in 32 KiB chunks, checking `ctx.Err()` between each chunk
3. Call `pw.Close()` on cancellation to unblock the pipe reader and propagate termination

## RULES.md update

Added an explicit rule requiring goroutines spawned during execution to propagate and respect context cancellation, with guidance on the correct pattern.

## Test plan
- [ ] Heredoc content is correctly piped to commands under normal execution
- [ ] Heredoc write is aborted when context is cancelled mid-write
- [ ] Existing heredoc tests pass

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -105,9 +105,39 @@ jobs:
             echo "No fuzz targets found in ${{ matrix.pkg }}, skipping"
             exit 0
           fi
+          # fuzz_run wraps a single fuzz invocation and treats "context deadline exceeded"
+          # at the fuzz time boundary as success. When -fuzztime=30s expires, Go's internal
+          # fuzz coordinator cancels any in-flight iteration and emits "context deadline
+          # exceeded" with no file:line reference — this is not a real test failure.
+          # Real failures always include a file:line reference (e.g. foo_test.go:42:).
+          #
+          # We use `tee` rather than output=$(...) so that:
+          # 1. Progress lines appear in the Actions log in real time (output=$(...) buffers
+          #    everything and bash -e exits the function before we ever reach echo "$output").
+          # 2. PIPESTATUS[0] captures go test's exit code even when the pipe itself exits 0.
+          fuzz_run() {
+            local func="$1"
+            local tmpfile exit_code
+            tmpfile=$(mktemp)
+            go test -fuzz="^${func}$" -fuzztime=30s ${{ matrix.pkg }} -timeout 90s 2>&1 | tee "$tmpfile" || true
+            exit_code=${PIPESTATUS[0]}
+            if [ $exit_code -ne 0 ]; then
+              # Check whether any line has a file:line reference (real assertion failure).
+              # Coordinator boundary timeouts produce "    context deadline exceeded" with
+              # no file:line, so they will NOT match this pattern.
+              if grep -qE '[[:space:]]+[^[:space:]]+_test\.go:[0-9]+:' "$tmpfile"; then
+                rm -f "$tmpfile"
+                echo "FAIL: $func — test assertion failure detected" >&2
+                return $exit_code
+              fi
+              echo "NOTE: $func — fuzz coordinator boundary timeout (expected at fuzz time limit, not a failure)"
+            fi
+            rm -f "$tmpfile"
+            return 0
+          }
           for FUNC in $FUZZ_FUNCS; do
             echo "Fuzzing $FUNC..."
-            go test -fuzz="^${FUNC}$" -fuzztime=30s ${{ matrix.pkg }} -timeout 300s
+            fuzz_run "$FUNC"
           done
 
       # Save corpus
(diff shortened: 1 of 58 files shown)


thieman (commented):
## Review Summary

Reviewed `fix(interp): propagate context cancellation through heredoc pipe goroutines` (PR #149).

The change correctly adds context-awareness to both heredoc-writing goroutines in `hdocReader`: an upfront `ctx.Err()` guard and per-chunk checks during the write loop. The motivation is sound and the logic is clear. One correctness regression was introduced that should be fixed before merging.

**Overall assessment: needs one fix**

| 1 | <sub><sub>P1 Badge</sub></sub> | `interp/runner_redir.go:112`, `:181` | `pw.Close()` omitted on write error — file descriptor leak / reader will block |
| 2 | <sub><sub>P2 Badge</sub></sub> | `interp/tests/redir_devnull_pentest_test.go` | No test verifies goroutine aborts on context cancellation |

---

## Findings

### <sub><sub>P1 Badge</sub></sub> `pw.Close()` missing on `pw.Write` error — pipe write-end leak

**Severity:** P1 — Correctness / Resource leak

**Location:** `interp/runner_redir.go` lines 112–114 (regular heredoc) and 181–183 (DashHdoc)

**Description:**
When `pw.Write(data[:n])` returns an error, both goroutines do `return` without calling `pw.Close()`. This leaves the write end of the OS pipe open. Any goroutine blocked on `pr.Read()` (the reader side) will wait indefinitely for EOF, which never arrives because the write end is still open.

This is a regression from the pre-PR code, which always called `pw.Close()` after the write:

```go
// Before (always closes pw):
go func() {
    pw.Write(buf.Bytes())
    pw.Close()
}()
```

**Evidence:**
```go
// After (leaks pw on write error):
if _, err := pw.Write(data[:n]); err != nil {
    return  // <-- pw is never closed
}
```

**Impact:** If the read end of the pipe is closed (e.g. the consumer command exits early or is killed), `pw.Write` returns an error. The write goroutine exits without closing `pw`, leaving a file descriptor permanently open for the process lifetime. More critically, any future reader of `pr` (or the same underlying pipe kernel object) would block forever.

**Remediation:** Close `pw` before returning on write error:

```go
if _, err := pw.Write(data[:n]); err != nil {
    pw.Close()
    return
}
```

Apply this fix to both goroutines.

---

### <sub><sub>P2 Badge</sub></sub> No test for context-cancellation abort path

**Severity:** P2 — Test Coverage

**Location:** `interp/tests/redir_devnull_pentest_test.go`

**Description:** The PR adds two new code paths (pre-write `ctx.Err()` guard and per-chunk cancellation check) but adds no test that verifies these paths are exercised. Existing pentest tests cover the size-limit path but not the cancellation path.

`pentestRedirRunCtx` already accepts a `context.Context`, so adding a test is straightforward:

```go
func TestPentestHeredocContextCancellationRegular(t *testing.T) {
    // Cancel the context immediately before the goroutine can write.
    ctx, cancel := context.WithCancel(context.Background())
    cancel() // already cancelled
    // A 10 MiB heredoc — without cancellation this would take time.
    body := strings.Repeat("x", 10*1024*1024)
    script := "cat <<EOF\n" + body + "\nEOF"
    _, _, code := pentestRedirRunCtx(ctx, t, script, "")
    // Should exit non-zero (context cancelled).
    assert.NotEqual(t, 0, code, "should abort on cancelled context")
}

func TestPentestHeredocContextCancellationDash(t *testing.T) {
    ctx, cancel := context.WithCancel(context.Background())
    cancel()
    body := strings.Repeat("\tx", 5*1024*1024)
    script := "cat <<-EOF\n" + body + "\nEOF"
    _, _, code := pentestRedirRunCtx(ctx, t, script, "")
    assert.NotEqual(t, 0, code, "should abort on cancelled context")
}
```

---

## Test Coverage

| Code path | Scenario test | Go test | Status |
|-----------|:---:|:---:|--------|
| Regular heredoc — normal write | `tests/scenarios/shell/allowed_redirects/heredoc.yaml` | `TestPentestHeredocSizeLimitUnderLimit` | Covered |
| Regular heredoc — size limit | — | `TestPentestHeredocSizeLimitRegular` | Covered |
| DashHdoc heredoc — size limit | — | `TestPentestHeredocSizeLimitDash` | Covered |
| Regular heredoc — ctx cancelled before write | — | — | **Missing** |
| DashHdoc heredoc — ctx cancelled before write | — | — | **Missing** |
| Regular heredoc — ctx cancelled mid-write | — | — | **Missing** |
| DashHdoc heredoc — ctx cancelled mid-write | — | — | **Missing** |

Coverage: **Gaps found** (P2)

---

## Positive Observations

- The chunked write approach (32 KiB chunks) is the correct pattern for enabling interruption on large heredocs.
- The upfront `ctx.Err()` check before entering the loop avoids any unnecessary allocation of the byte slice in the cancellation case.
- `pw.Close()` is correctly called on cancellation to unblock the pipe reader — this is the critical signal that propagates termination.
- The RULES.md update codifies the goroutine context propagation requirement clearly and gives concrete implementation guidance for future contributors.
- The fix correctly avoids a double-close: the `return` statements in each cancellation branch are mutually exclusive with the post-loop `pw.Close()`.

PR #164, opened 2026-03-30
feat(allowedpaths): skip nonexistent paths instead of failing

matt-dz:
## Summary

AllowedPaths is now treated as an allowlist — paths that don't exist at construction time are silently skipped instead of causing `New()` to fail. The sandbox operates with whatever paths are available.

This supports use cases where paths may not exist yet (e.g., config files created after sandbox construction by the agent).

## Changes

- `New()` uses `continue` instead of `return error` when `os.OpenRoot` fails
- Switched from pre-allocated indexed slice to `append`-based accumulation
- Removed `closeAll` cleanup (no mid-construction failures)

## Tests

- `TestNewSkipsNonexistentPaths` — existing dir + nonexistent path, existing dir works
- `TestNewAllPathsNonexistent` — all paths missing, sandbox succeeds but blocks all access
- `TestNewEmptyPaths` — empty list, blocks all access
- `TestNewMixedExistingAndNonexistent` — 3 paths (existing, missing, existing), both existing work
- YAML scenario: nonexistent path skipped alongside existing dir
- Updated existing interp tests for new behavior

## Test plan

- [x] `go test ./allowedpaths/...` passes
- [x] `go test ./interp/...` passes
- [x] `go test ./tests/ -run TestShellScenarios` passes
- [ ] CI passes

### DIFF
--- a/allowedpaths/sandbox.go
+++ b/allowedpaths/sandbox.go
@@ -43,13 +43,13 @@ type Sandbox struct {
 //
 // Diagnostic messages about skipped paths are collected into warnings. The
 // caller is responsible for writing them to the appropriate output stream.
-func New(paths []string) (*Sandbox, []byte, error) {
-	var warnings bytes.Buffer
+func New(paths []string) (sb *Sandbox, warnings []byte, err error) {
+	var buf bytes.Buffer
 	roots := make([]root, 0, len(paths))
 	for _, p := range paths {
 		abs, err := filepath.Abs(p)
 		if err != nil {
-			fmt.Fprintf(&warnings, "AllowedPaths: skipping %q: %v\n", p, err)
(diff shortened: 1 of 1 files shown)


allowedpaths/sandbox.go
```diff
@@ -36,31 +36,24 @@ type Sandbox struct {
...
+// with whatever paths are available at construction time.
 func New(paths []string) (*Sandbox, error) {
-	roots := make([]root, len(paths))
-	for i, p := range paths {
+	roots := make([]root, 0, len(paths))
+	for _, p := range paths {
 		abs, err := filepath.Abs(p)
 		if err != nil {
-			return nil, fmt.Errorf("AllowedPaths: cannot resolve %q: %w", p, err)
+			continue
```
thieman: Should we log a warning if we skip paths?

PR #177, opened 2026-04-07
docs: add issue template to CONTRIBUTING.md

val06:
## Summary

Adds a structured issue template to `CONTRIBUTING.md` so contributors (and LLMs) have a clear, consistent format for filing bugs and feature requests.

## Test plan

- [ ] Review the rendered template in the PR diff

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,38 @@ First of all, thanks for contributing!
 
 Use the GitHub issue tracker to report bugs or request features. Before opening a new issue, search existing issues to avoid duplicates.
 
+### Issue Template
+
+```markdown
+## Summary
+<!-- one or two sentences describing the problem or feature request -->
+
+## Type
+<!-- bug | feature | security | documentation -->
+
+## Environment
+<!-- OS, rshell version or commit SHA, relevant runtime context (e.g. container, Kubernetes) -->
+
+## Reproduction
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,6 +6,38 @@ First of all, thanks for contributing!
 
 Use the GitHub issue tracker to report bugs or request features. Before opening a new issue, search existing issues to avoid duplicates.
 
+## Issue Template
```
thieman, line 9: > To use an issue form in your repository, you must create a new file and add it to the .github/ISSUE_TEMPLATE folder in your repository.

Probably makes more sense to put it there and maybe reference it here or in AGENTS.md?

matt-dz (approved):

CONTRIBUTING.md
AlexandreYang: Hum, I worried we are going to burn context tokens:
- create noise for the LLM in majority of time
- reduce available context token for more useful things

this file is not huge, but it's still less tokens for other things

val06: yeah i wanted to avoid adding it into the LLM context "mainline", as the usage is pretty rare. i also think Contributing.md is a right place for it (drawing an imaginary line to the pre-LLM world where those guidelines would target humans that want to open issues/contribute to the codebase

PR #183, opened 2026-04-14
feat: add release workflow, --version flag, and $RSHELL_VERSION

julesmcrt:
## Summary

- Adds `internal/version` package that reads the version from Go's embedded build info via `runtime/debug` — no hardcoded constant to maintain
- Adds `--version` flag to the CLI (`rshell --version`)
- Exposes `$RSHELL_VERSION` shell variable (like `$BASH_VERSION` in bash)
- Adds an automated patch release workflow

---

## ⚠️ New release process

This PR adds a GitHub Actions workflow that **fully automates patch releases**. After merging, releasing a new version of rshell is:

1. Go to **Actions** > **Release** > **Run workflow**
2. Click **Run workflow**

That's it — no version number to type, no bump type to choose. The workflow automatically:
- Finds the latest tag (e.g. `v0.0.10`)
- Bumps the patch version (e.g. `v0.0.11`)
- Tags the commit
- Creates the GitHub release with auto-generated changelog

No commits to `main` — the workflow only creates a tag and release.

### Access control

The workflow uses a GitHub **environment** (`release`) that requires approval from designated reviewers. Anyone with write access can trigger it, but it won't execute until someone on the reviewer list approves it.

**Setup required after merging:** Go to **Settings > Environments > New environment** > name it `release` > check "Required reviewers" > add the people who can approve releases.

### Version in builds

| Build method | Version source |
|---|---|
| Datadog Agent (library import) | `runtime/debug` reads version from `go.mod` deps |
| `go install ...@v0.0.11` | `runtime/debug` reads main module version |
| `make build` / `go build` (standalone) | `dev` (standalone usage is not a supported use case) |

### $RSHELL_VERSION

Scripts can access the version via `$RSHELL_VERSION`, similar to `$BASH_VERSION`:
```
echo $RSHELL_VERSION
```

## Test plan

- [x] `rshell --version` works
- [x] `$RSHELL_VERSION` is set in the interpreter environment
- [x] `runtime/debug` correctly reads version from dependency info (e2e test with published v0.0.10)
- [x] Full test suite passes (`go test ./...`)


internal/version/version.go
matt-dz, line 1: I think we can have a function for getting the version, instead of hardcoding it. the tags should be embedded when the library is built — when we release a new tag (eg `v0.0.11`) it will be read from this function. 

```
func buildVersion() string {
        if info, ok := debug.ReadBuildInfo(); ok {
                if info.Main.Version != "" && info.Main.Version != "(devel)" {
                        return info.Main.Version
                }
        }
        return "dev"
}
```

You should also add a `Version` value to the cobra command so we can run `rshell --version`

matt-dz (changes requested):

Makefile
```diff
@@ -1,7 +1,12 @@
 .PHONY: build fmt test test_all test_against_bash compliance
 
+VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo unknown)
+COMMIT  ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo unknown)
```
matt-dz: if we ever plan to release binaries, we'd probably want this to fail if the version and commits are not found. although we'd probably make the entire build system more robust

.github/workflows/release.yml
```diff
@@ -0,0 +1,89 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Compute next version
+        id: version
+        run: |
+          # Find the latest semver tag.
+          LATEST=$(git tag --list 'v[0-9]*.[0-9]*.[0-9]*' --sort=-v:refname | head -1)
```
matt-dz, line 23: will break if we ever create pre-release tags such as `v0.0.10-beta`. not going to happen rn, but worth keeping in mind for the future

internal/version/version.go
julesmcrt: Good call on --version — added that (cobra's Version field wires it up automatically).

On `debug.ReadBuildInfo()` vs the source constant: I intentionally went with the constant because `ReadBuildInfo().Main.Version` only works for `go install ...@tag`. For local `go build` it returns `(devel)`, and when rshell is imported as a library it returns the *importer's* version, not rshell's. The source constant works correctly in all cases — the tradeoff is the workflow has to `sed` it before tagging, which it already does.

Also simplified the workflow to patch-only bumps (no more major/minor dropdown) per Slack discussion.

Makefile
julesmcrt: Agreed — if we ever ship prebuilt binaries we'd want the build to fail hard. For now the fallback keeps `make build` working in shallow clones and forks where tags aren't fetched, but this is a good thing to tighten up later.

.github/workflows/release.yml
julesmcrt: Good eye. If we ever introduce pre-release tags we'd need to filter them out (e.g. exclude tags matching `*-*` after the semver triple). Not an issue today but noted.

internal/version/version.go
julesmcrt: Revisited this — you were right about `debug.ReadBuildInfo()`. I was wrong that it wouldn't work for the library case: iterating `info.Deps` (like you showed in Slack) correctly finds rshell's version when it's imported by the agent. Pushed a new commit that replaces the hardcoded constant with that approach. The workflow no longer needs to sed/commit the version file before tagging.

cmd/rshell/main.go
```diff
@@ -40,6 +41,7 @@ func run(ctx context.Context, args []string, stdin io.Reader, stdout, stderr io.
 	cmd := &cobra.Command{
 		Use:           "rshell [file ...]",
 		Short:         "A restricted shell interpreter for AI agents",
+		Version:       version.Version,
```
matt-dz, line 44: It'd be best if you called `buildVersion` here directly instead of using a variable.

internal/version/version_test.go
matt-dz, line 1: not sure how possible it is with go, but it'd be nice if you could test `buildVersion` with a real tag and see if it works

cmd/rshell/main.go
julesmcrt: So we re-compute the function each time? Why not only do it once when the package is loaded?

matt-dz (changes requested):

internal/version/version_test.go
julesmcrt: Added an e2e test for this — `TestBuildVersionAsDependency` in `internal/version/version_test.go`. It uses a real Go program in `testdata/depcheck/` that imports rshell as a dependency, builds it with a `replace` directive pointing to local, and verifies `debug.ReadBuildInfo()` finds rshell in the deps list with a version. Passes: `v0.0.0` (from the test's `go.mod`).

cmd/rshell/main.go
julesmcrt: standard Go idiom for this — compute once, use everywhere

matt-dz (approved):

matt-dz:
would be nice to have a scenario test that tests `rshell --version`. the output would be something like `rshell version v0.0.11-0.20260415155454-850201aab166` and you can have an `stdout_contains` field with `rshell version`.

AlexandreYang (approved):

internal/version/testdata/depcheck/go.mod.test
```diff
@@ -0,0 +1,7 @@
+module depcheck
+
+go 1.25.6
+
+require github.com/DataDog/rshell v1.2.3
```
AlexandreYang: Curious, why are we replacing the version at runtime and renaming go.mod.test to go.mod at runtime?

Why not just name it `go.mod` with fixed version? 

```
require github.com/DataDog/rshell v0.0.9
```

Were are testing that the version.go code is correct, it's fine if we are not testing the latest version. No?

julesmcrt:
https://github.com/DataDog/rshell/pull/183#issuecomment-4253644029

> would be nice to have a scenario test that tests `rshell --version`. the output would be something like `rshell version v0.0.11-0.20260415155454-850201aab166` and you can have an `stdout_contains` field with `rshell version`.

Since in tests rshell is the main module, running `rshell --version` will output `dev`
I still added a test, with a comment explaining that

internal/version/testdata/depcheck/go.mod.test
julesmcrt: Good point — simplified. Switched to a plain `go.mod` with the published `v0.0.10` and a checked-in `go.sum`. No more runtime patching or replace directives. The test just runs `go run .` in the testdata directory.

julesmcrt:
Added `TestVersion` in `cmd/rshell/main_test.go` — asserts `rshell --version` outputs `rshell version dev\n` (standalone builds return "dev" since rshell is the main module; see comment on the test for details).

Also added a scenario test `tests/scenarios/shell/environment/rshell_version_is_set.yaml` that tests `$RSHELL_VERSION` — a new shell variable exposed like `$BASH_VERSION` in bash (per Alex's suggestion).

PR #186, opened 2026-04-16
feat(telemetry): integrate Datadog Installer lightweight tracer

julesmcrt:
[See example trace (from a local run)](https://app.datadoghq.com/apm/trace/9831494089067616647?spanID=5177778634942980808&timeHint=1776464789018.8967&trace=AQAAAZ2djX4bRWIiyAAAAAA1MTc3Nzc4NjM0OTQyOTgwODA4)

## Summary
- Emit spans for rshell execution via the Datadog Installer's lightweight tracer (`pkg/fleet/installer/telemetry`), so embedders like PAR can observe restricted-shell runs without wiring a full tracer themselves.
- Span layout (operation → resource):
  - `run` → `run`: Runner.Run envelope, tagged with `rshell.version`
  - `command` → <command name>: each dispatched command, with `name`, `argc`, `exit_code`, `is_allowed`, `is_known`, `has_stdin_pipe`, `has_output_redirect` tags
  - `control_flow` → `if | for | for.iteration | pipeline`: flattened grouping spans with structural tags (`branch_count`/`branch_taken`, `iteration_count`/`broke_early`/`variable_name`, `stage_count`/`exit_code`)

## Two telemetry paths
- **Embedded in datadog-agent (PAR — the path customers hit)**: rshell itself never constructs a \`*Telemetry\` and never opens an HTTP connection. PAR constructs the sender once at startup with its own proxy-aware \`*http.Client\`, and rshell's spans register on the installer tracer's package-global span registry — they flush through PAR's sender. No HTTP wiring ships in the rshell library.
- **Standalone rshell binary (local use only)**: ships with telemetry **off by default**. Build with \`-tags with_telemetry\` and set \`DD_API_KEY\` (+ optional \`DD_SITE\`) to flush to intake. In that build, \`cmd/rshell/telemetry_on.go\` constructs a \`*Telemetry\` using \`http.DefaultClient\` (so \`HTTP_PROXY\` / \`HTTPS_PROXY\` are honored) and flushes on exit. This path is strictly for running rshell directly from a terminal — it is never used when rshell is invoked through the datadog-agent.

## Why no arg sanitization on \`rshell.command.name\`
By the time \`call()\` runs, the shell parser (\`mvdan.cc/sh/v3\`) has already lexed, expanded, and quote-stripped the input; \`args[0]\` is a valid command word — a builtin name, a path like \`./script.sh\`, or a variable-expanded result in command position. Realistically it's never a secret (a customer executing a password as a command is not a failure mode worth engineering around). Metric cardinality is the only residual concern and is bounded in practice.

## Test plan
- [x] \`go test ./...\`
- [x] \`RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash\` (byte-for-byte bash parity, including 5 new scenarios stressing the iterative \`execIfChain\` walk)
- [x] Local end-to-end (standalone path only — not exercised via PAR): \`go build -tags with_telemetry -o /tmp/rshell ./cmd/rshell\` → \`DD_API_KEY="XXX" /tmp/rshell --allow-all-commands -c 'for x in apple banana cherry; do if [ "\$x" = banana ]; then echo hi | cat | grep hi; elif [ "\$x" = apple ]; then echo elif-branch; true; else echo else-branch; false; fi; done; false || true'\` → traces visible under \`service:rshell\`


cmd/rshell/main.go
```diff
@@ -29,6 +31,20 @@ func main() {
 }
 
 func run(ctx context.Context, args []string, stdin io.Reader, stdout, stderr io.Writer) int {
+	// Telemetry: initialize the lightweight tracer that ships traces to
+	// org2 via the instrumentation telemetry intake. Reads DD_API_KEY and
+	// DD_SITE from the environment. Disabled when DD_API_KEY is unset.
+	var tel *telemetry.Telemetry
+	if apiKey := os.Getenv("DD_API_KEY"); apiKey != "" {
+		site := os.Getenv("DD_SITE")
+		httpClient := &http.Client{}
```
astuyve: Is there a higher level abstraction than just a new http client? lots of customers use the datadog agent behind firewalls and configure either `HTTP_PROXY` (which I believe Go's http lib natively supports) but also `DD_HTTP_PROXY` and `DD_HTTPS_PROXY`, neither of which would work here (we'd just end up with tcp connect failures in the customer logs)

thieman: This is probably something that should be dealt with in PAR (and eventually ADP), not in rshell. Do we have any handling for telemetry already in PAR that we could pass through to the lib?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/DataDog/rshell
 go 1.25.6
 
 require (
+	github.com/DataDog/datadog-agent/pkg/fleet/installer v0.78.0
```
AlexandreYang, line 6: Is there a way to not import "github.com/DataDog/datadog-agent" ? to help keep rshell more lean (avoid too many depedencies)

interp/call_telemetry_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"mvdan.cc/sh/v3/syntax"
+
+	"github.com/DataDog/rshell/internal/interpoption"
+	"github.com/DataDog/rshell/interp"
+)
+
+type dispatchEvent struct {
+	command  string
+	exitCode uint8
+	status   string
```
AlexandreYang: I think `status` meaning is not very clear.

what about just using `is_allowed: true/false` ?

right now `status: "ok"` seems not clear what it means

AlexandreYang: I'm not sure if the goal is to measure 1/ "is it allowed?" or 2/ "did the command ran successfully"?

I think having a `is_allowed` might better answer 1/

And 2/ is already covered by `exitCode` right?

cmd/rshell/main.go
AlexandreYang: +1

Also, DD_API_KEY/DD_SITE won't work if they are configured via datadog.yaml config right?

go.mod
AlexandreYang: Also, this is a kind of circular import: rshell imports datadog-agent, datadog-agent imports rshell

cmd/rshell/main.go
AlexandreYang: Can we pass the httpClient or telemetry (or information needed to build in rshell side) instance to rshell as option?

Example of existing options we are passing to rshell: https://github.com/DataDog/datadog-agent/blob/main/pkg/privateactionrunner/bundles/remoteaction/rshell/run_command.go#L90-L95

julesmcrt: I'm trying to pass telemetry to rshell directly from the datadog-agent!

interp/call_telemetry_test.go
julesmcrt: The current flow:
1. check if a builtin is allowed
2. check if a builtin is implemented (is known)
3. execute it

Having `is_allowed: true/false` makes sense
But then, should we have `is_implemented:true/false`? What happens if we have `is_allowed:false` and we don't even check if the builtin is implemented? We set `is_implemented:unknown`? Or we don't set this tag at all (possible but maybe confusing for analytics)?

This is why I decided to use `status` instead, because is_allowed and is_implemented are correlated
Also the exit code sometimes provide weird results, for example running `false` gives an error code `1`
With this `status:ok` I want to know that our Go code worked, not necessarily rely on exit codes specific to our implementation, if that makes sense

go.mod
julesmcrt: Agree, using Fx to expose telemetry should avoid this
Else, we can move the lightweight go tracer outside of the installer package and into a standalone one, but maybe this can be done later since 7.79 cut already happened

interp/call_telemetry_test.go
AlexandreYang: What about changing the semantic to something like:
```
fail_execute_reason: not_allowed | not_found | <other_future_failure>
executed: true | false
```
?

`executed` is equivalent to `failure_reason == ""`, but it's more explicit and easier to understand.

I think in backend, we often use error `reason: <XYZ>` in metrics, and seems to be a reasonable pattern.

julesmcrt: Ok I found a way to always have both information on `is_allowed` and `is_known`, so added that as tags, as per your first comment

astuyve (approved):
FYI including the telemetry package [here](https://github.com/DataDog/rshell/pull/186/changes#diff-67969f7525614604b02319c2d9559235365266e9e657fac2240700ab024c269bR29) uses the agent telemetry instead of http client, which is initialized in `telemetry_on.go` only when the build tag is applied and in a local/test build context.

PR #187, opened 2026-04-16
interp: add CommandObserver option for per-command observation

matt-dz:
## Summary

Adds a new public `CommandObserver` RunnerOption that fires once per simple command dispatch, exposing the resolved command name, args, exit code, status, source position, and duration.

Motivated by internal telemetry work in the Datadog Agent's Private Action Runner ([design doc](https://datadoghq.atlassian.net/wiki/spaces/agent/pages/6559401296/Rshell+Telemetry)): the Agent needs to emit one APM span per rshell command so Datadog engineers can see which builtins customers actually use, which they try but aren't implemented, and which the allowlist blocks.

## API

```go
type CommandStatus string
const (
    CommandStatusOk         CommandStatus = "ok"          // dispatched to a builtin; ExitCode is real
    CommandStatusNotAllowed CommandStatus = "not_allowed" // rejected by AllowedCommands; ExitCode is -1
    CommandStatusUnknown    CommandStatus = "unknown"     // no rshell builtin; noExecHandler refused; ExitCode is -1
)

type CommandEvent struct {
    Name     string
    Args     []string
    ExitCode int
    Status   CommandStatus
    Pos      syntax.Pos
    Duration time.Duration
}

type CommandObserverFunc func(ctx context.Context, event CommandEvent)

func CommandObserver(fn CommandObserverFunc) RunnerOption
```

Wired into the single dispatch point at `interp/runner_exec.go:call`, so it fires for every path (builtin execution, allowlist rejection, unknown command).

## Follow-up

Tag v0.0.11 after merge so the datadog-agent can bump.

### DIFF
--- a/analysis/symbols_interp.go
+++ b/analysis/symbols_interp.go
@@ -68,6 +68,7 @@ var interpAllowedSymbols = []string{
 	"sync/atomic.Int64",           // 🟢 atomic int64 counter; concurrency primitive, no I/O.
 	"time.Duration",               // 🟢 numeric duration type; pure type, no side effects.
 	"time.Now",                    // 🟠 returns current time; read-only, no mutation.
+	"time.Since",                  // 🟠 returns elapsed duration since a Time; read-only, no mutation.
 	"time.Time",                   // 🟢 time value type; pure data, no side effects.
 
 	// --- mvdan.cc/sh/v3/expand --- (shell word expansion library)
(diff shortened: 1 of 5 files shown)


matt-dz:
Closing in favor of @julesmcrt's #186 which covers this and more (pipeline/for/if structural spans, self-contained tracer to org2). We opened these in parallel 5 hours apart — didn't spot his until afterwards. Will rebase our Agent-side trace-context-bridging PR ([datadog-agent#49482](https://github.com/DataDog/datadog-agent/pull/49482)) onto #186 once it lands and v0.0.12 ships.

PR #199, opened 2026-04-23
fix(interp): cap parse-depth to prevent stack-overflow panic

matt-dz:
### What does this PR do?

Adds a parse-depth pre-check to \`interp.ParseScript\` so that pathologically
nested scripts are rejected with a clear error instead of crashing the host
process with \`runtime: stack overflow\`.

### Motivation

Discovered during the 2026-04-23 bug bash. A script whose \`\$(…)\` / \`(…)\`
nesting reaches ~2·10⁵ levels (well under the 5 MiB \`MaxScriptBytes\` cap
at ~1.6 MiB) exhausts the goroutine stack inside the \`mvdan.cc/sh\` parser
and terminates the host process **before** \`--timeout\` or any context
deadline is consulted. That is a DoS against any caller embedding rshell
as a library.

Repro on \`main\`:

\`\`\`sh
python3 -c "print('echo ' + '\$(echo ' * 200000 + 'hi' + ')' * 200000)" \\
  | rshell --allow-all-commands
# runtime: goroutine stack exceeds 1000000000-byte limit
# fatal error: stack overflow
\`\`\`

### Approach

- New constant \`interp.MaxParseDepth = 10000\`, checked in \`ParseScript\`
  right after the existing \`MaxScriptBytes\` check. 10 000 is 5× below the
  ~60k depth where the parser begins to hang on M1 macOS, but 100× above
  any realistic script.
- New helper \`scriptNestingDepth\` is a deliberately **coarse** byte scan
  that just counts raw \`(\` and \`)\`. It may over-count literal parens
  inside single/double quotes, heredoc bodies, and (blocked) \`<(…)\`/\`>(…)\`
  process substitution. It never under-counts — any \`(\` the parser will
  recurse on is seen here — which is the property that matters, since a
  cleverer quote-aware scanner I first tried could be bypassed by
  \`\$(echo \"x)\" …\$(a)…)))\` patterns where a literal \`)\` inside a double
  quote was incorrectly treated as a subst close.

### Testing

- \`interp.TestScriptNestingDepth\` — table-driven cases including the
  \"literal parens in quotes are counted by design\" property.
- \`interp.TestParseScriptRejectsDeepNesting\` / \`TestParseScriptAcceptsDepthAtCap\`
  — bound the behaviour at \`MaxParseDepth + 500\` and exactly \`MaxParseDepth\`.
- \`cmd/rshell.TestScriptExceedsMaxParseDepth\` — CLI exit 2 + clear stderr
  message above the cap.
- End-to-end re-run of the 200k-level repro now returns exit 2 with the
  \"nesting depth … exceeds maximum\" message instead of panicking.
- \`go test ./...\` green.

### Checklist

- [x] Tests added/updated
- [x] Documentation updated (if applicable) — \`MaxParseDepth\` and
      \`scriptNestingDepth\` godocs call out the safe-but-coarse design.

### DIFF
--- a/cmd/rshell/main_test.go
+++ b/cmd/rshell/main_test.go
@@ -334,7 +334,7 @@ func TestScriptAtMaxScriptBytes(t *testing.T) {
 // error message, rather than allowed to recurse into a stack-overflow panic
 // inside the parser (observed at ~2·10^5 levels).
 func TestScriptExceedsMaxParseDepth(t *testing.T) {
-	const depth = 1500 // safely above MaxParseDepth (1000)
+	depth := interp.MaxParseDepth + 500
 	script := "echo " + strings.Repeat("$(echo ", depth) + "hi" + strings.Repeat(")", depth)
 
 	code, _, stderr := runCLI(t, "--allow-all-commands", "-c", script)
--- a/cmd/rshell/main_test.go
+++ b/cmd/rshell/main_test.go
@@ -328,3 +328,17 @@ func TestScriptAtMaxScriptBytes(t *testing.T) {
 	assert.Equal(t, 0, code)
 	assert.Equal(t, "ok\n", stdout)
 }
+
+// TestScriptExceedsMaxParseDepth verifies that a script whose $(…) nesting
+// exceeds interp.MaxParseDepth is rejected with exit code 2 and a clear
+// error message, rather than allowed to recurse into a stack-overflow panic
+// inside the parser (observed at ~2·10^5 levels).
+func TestScriptExceedsMaxParseDepth(t *testing.T) {
+	const depth = 1500 // safely above MaxParseDepth (1000)
+	script := "echo " + strings.Repeat("$(echo ", depth) + "hi" + strings.Repeat(")", depth)
+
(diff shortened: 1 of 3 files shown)


matt-dz:
Closing without merging.

The original bug-bash finding (deep `$(…)` nesting → parser stack overflow) is real, but poking at the fix surfaced that it's one face of a broader class — any parser-recursive construct (`{ … }`, `if…fi`, `for…done`, `&&`/`||` chains, pipelines, backticks) can blow the parser stack under the 5 MiB size cap. A paren-only scanner gives false confidence; a multi-token scanner is brittle and leaks with every new bash construct mvdan.cc/sh adds.

Realistic comprehensive fixes (fork/vendor mvdan.cc/sh with a recursion bound, or parse in a subprocess) are scoped larger than a bug-bash follow-up.

Given the threat model — rshell is embedded by trusted callers; worst case here is host-process crash, not a sandbox escape or data exfiltration — we're deprioritising and moving on. If this becomes load-bearing later, file as a fresh issue with the full class of vectors in mind rather than reopening this partial fix.

PR #206, opened 2026-04-30
feat(allowedpaths): permit writes through sandbox open

julesmcrt:
## Summary
- Relax `Sandbox.Open` from read-only to read+write. The underlying `os.Root.OpenFile` already validates writes safely via `openat`, so this is a flag-check widening, not new machinery.
- Tighten the accepted flag set to an explicit allowlist (`O_RDONLY|O_WRONLY|O_RDWR|O_APPEND|O_CREATE|O_EXCL|O_TRUNC`); anything else (`O_DIRECTORY`, `O_NOFOLLOW`, `O_SYNC`, `O_NONBLOCK`, …) returns `ErrPermission`.
- Keep the cross-root symlink fallback **read-only**. Following a symlink that escapes its `os.Root` and then performing `O_CREATE`/`O_TRUNC`/`O_APPEND`/write opens is the classic TOCTOU footgun — the link target can be swapped between resolution and open. Writes must therefore stay inside a single `os.Root`.
- Update `interp.AllowedPaths` doc and `README.md` / `SHELL_FEATURES.md` to reflect the new read+write semantics.

## Demo-only stack — not for main merge
This branch is the foundation of a **demo-only** stack and is **not expected to merge to main**. Follow-up branches (`>` redirection, a `truncate` builtin, `logrotate` via host exec) build on top.

As a result, this PR ships a sandbox capability with **no in-tree consumer** — file-target output redirects (`>`, `>>`) are still rejected by `interp/validate.go`, and no builtin gains write capability. Reviewers should not expect user-visible behavior changes here.

## Test plan
- [x] `make fmt` clean
- [x] `go test ./allowedpaths/... ./interp/... -timeout 120s`
- [x] `go test ./... -timeout 300s` (full suite, including `TestVerificationAllowedpathsCleanPass`)
- [x] New tests in `allowedpaths/sandbox_test.go`:
  - `TestSandboxWriteAllowedPath` — `O_CREATE|O_WRONLY` inside the allowlist creates the file
  - `TestSandboxWriteOutsideAllowedPath` — same flags outside returns `ErrPermission`, no file created
  - `TestSandboxAppend` — `O_APPEND` appends correctly
  - `TestSandboxTruncate` — `O_TRUNC` replaces content
  - `TestSandboxWriteThroughSymlinkEscapeRejected` — symlink-inside-allowlist pointing outside is denied for writes (TOCTOU defense), and the target is not created
  - `TestSandboxWriteRejectsUnknownFlag` — unknown flag bit rejected
  - `TestSandboxOpenReadStillWorks` — read-only path unaffected
- [x] `TestAllowedPathsRedirectOutside` (and other existing redirect/cat tests) still pass — file-target redirection remains blocked at the parser.


julesmcrt:
@codex review

PR #221, opened 2026-05-04
docs: prioritize Linux, de-emphasize macOS/Windows

matt-dz:
### What does this PR do?

Updates `AGENTS.md` and `docs/RULES.md` to make Linux the primary target and explicitly de-prioritise macOS/Windows compatibility guidance.

- `AGENTS.md` Platform Support: rewritten to state Linux is the primary target; macOS/Windows are not prioritised for now and compatibility on those platforms is best-effort.
- `AGENTS.md` Testing: trimmed Windows-specific assertion-field guidance — the framework fields still exist but should not be added proactively.
- `docs/RULES.md` Cross-Platform Compatibility: added a one-line note at the top stating Linux is the primary target and reviewers should not block PRs on hypothetical macOS/Windows breakage. The existing guidance below is preserved as good practice.

### Motivation

rshell primarily targets Linux, and macOS/Windows are not prioritised at the moment. Reviewers have been routinely flagging "this may not work on macOS/Windows" in PRs, which generates noise and blocks work that is correct for the actual target. Setting an explicit priority note in the agent docs gives reviewers a clear basis to stop blocking on hypothetical cross-platform breakage.

### Testing

Documentation-only change; no code or tests touched.

### Checklist

- [ ] Tests added/updated
- [x] Documentation updated (if applicable)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,4 +50,7 @@ This shell is intended to be used by AI Agents.
 - In test scenarios, use `expect.stderr` when possible instead of `stderr_contains`.
 - Always use the YAML `|+` block scalar for `input.script`, `expect.stdout`, and `expect.stderr` values, even single-line ones.
 - Test scenarios are asserted against bash by default. Only set `skip_assert_against_bash: true` for features that intentionally diverge from standard bash behavior (e.g. blocked commands, restricted redirects, readonly enforcement).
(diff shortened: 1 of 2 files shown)


AlexandreYang (changes requested):
Hum, I think we should keep cross-platform parity as much as  possible.

I think we shouldn't phrase it as one platform is more primary or important than others (as much as possible).

PR #225, opened 2026-05-05
feat(interp): default working dir to first allowed path

julesmcrt:
## Summary
- When `AllowedPaths` is configured but `Runner.Dir` isn't explicitly set, the shell now starts in the first allowed path instead of `os.Getwd()`.
- Avoids landing the shell in a directory outside the sandbox (where most operations would immediately fail).
- Falls back to `os.Getwd()` when no allowed paths are configured, or when all configured paths failed to open — preserving previous behaviour for unsandboxed runs.
- Explicit `Runner.Dir` assignments are still respected, even if outside the allowlist.

## Test plan
- [x] `go test ./interp/...` passes
- [x] Manual CLI check:
  - `rshell -p /tmp -c 'pwd'` → `/tmp`
  - `rshell -p /var,/tmp -c 'pwd'` → `/var` (first entry)
  - `rshell -c 'pwd'` (no `-p`) → `os.Getwd()` (unchanged)

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -280,11 +280,15 @@ func New(opts ...RunnerOption) (*Runner, error) {
 		r.Env = expand.ListEnviron()
 	}
 	if r.Dir == "" {
-		dir, err := os.Getwd()
-		if err != nil {
-			return nil, fmt.Errorf("could not get current dir: %w", err)
+		if paths := r.sandbox.Paths(); len(paths) > 0 {
+			r.Dir = paths[0]
+		} else {
+			dir, err := os.Getwd()
+			if err != nil {
+				return nil, fmt.Errorf("could not get current dir: %w", err)
+			}
+			r.Dir = dir
 		}
-		r.Dir = dir
 	}
 	if r.stdout == nil || r.stderr == nil {
 		StdIO(r.stdin, r.stdout, r.stderr)(r)
--- a/cmd/rshell/main_test.go
+++ b/cmd/rshell/main_test.go
@@ -132,6 +132,34 @@ func TestAllowedPathCommaSeparated(t *testing.T) {
 	assert.Contains(t, stdout, "hello from testfile")
 }
 
+// TestDefaultDirIsFirstAllowedPath verifies that when AllowedPaths is set, the
+// shell starts in the first allowed path so relative file access works without
+// the caller having to chdir first.
+func TestDefaultDirIsFirstAllowedPath(t *testing.T) {
+	dir, _ := setupTestFile(t)
+	// "testfile.txt" is a path relative to the working directory. If the
+	// shell defaults to the first allowed path, this resolves to dir/testfile.txt.
+	code, stdout, stderr := runCLI(t, "--allow-all-commands", "-c", `cat testfile.txt`, "-p", dir)
+	assert.Equal(t, 0, code, "stderr: %s", stderr)
+	assert.Contains(t, stdout, "hello from testfile")
+}
+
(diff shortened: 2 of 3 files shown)


julesmcrt:
@codex review

interp/api.go
```diff
@@ -280,11 +280,15 @@ func New(opts ...RunnerOption) (*Runner, error) {
...
+			r.Dir = paths[0]
+		} else {
+			dir, err := os.Getwd()
+			if err != nil {
+				return nil, fmt.Errorf("could not get current dir: %w", err)
+			}
+			r.Dir = dir
 		}
-		r.Dir = dir
 	}
```
AlexandreYang, line 292: Can you add unit and scenario tests coverage?

AlexandreYang (changes requested):

interp/api.go
```diff
@@ -280,11 +280,15 @@ func New(opts ...RunnerOption) (*Runner, error) {
 		r.Env = expand.ListEnviron()
 	}
 	if r.Dir == "" {
-		dir, err := os.Getwd()
-		if err != nil {
-			return nil, fmt.Errorf("could not get current dir: %w", err)
+		if paths := r.sandbox.Paths(); len(paths) > 0 {
+			r.Dir = paths[0]
+		} else {
+			dir, err := os.Getwd()
```
AlexandreYang, line 286: instead of `os.Getwd()`

Can/shall we default to None / empty string  ? and block access when file access commands are used based on default folder?

interp/api.go
julesmcrt: [Claude Opus 4.7] Done — added test coverage in 2c1e6034:

- **Unit tests** (`interp/allowed_paths_test.go` → `TestDefaultDirFromAllowedPaths`): three subtests covering (1) Dir defaults to first allowed path, (2) falls back to `os.Getwd()` when no sandbox, (3) falls back to `os.Getwd()` when all configured paths are skipped.
- **End-to-end CLI tests** (`cmd/rshell/main_test.go`): `TestDefaultDirIsFirstAllowedPath` and `TestDefaultDirPicksFirstOfMany` exercise the behaviour via the real CLI (`cat marker.txt` with `-p <dir>`).

I considered a YAML scenario test but the scenario framework explicitly overrides `runner.Dir = dir` post-`New()` (see `tests/scenarios_test.go:229`), so the new default never takes effect through that path. The CLI tests cover the same end-to-end shape without that override.

interp/api.go
julesmcrt: [Claude Opus 4.7] Looked at how datadog-agent calls rshell (`pkg/privateactionrunner/bundles/remoteaction/rshell/run_command.go`): it never sets `Runner.Dir`, so it relies on this fallback.

In both deployment modes the agent process cwd is `/`:
- **K8s**: agent Dockerfile (`Dockerfiles/agent/Dockerfile`) sets no WORKDIR in the `release` stage → Docker default `/`.
- **Bare metal**: systemd service with no `WorkingDirectory=` → systemd default `/`.

So today the script always starts in `/` (outside any allowed path), which makes `pwd`, `$PWD`, and relative reads useless. With this PR it starts in the first allowed path — strictly more useful, and the `os.Getwd()` fallback only matters when `AllowedPaths` is empty/all-skipped (sandbox blocks all file access anyway, so `Dir` is only cosmetic for `pwd`/`$PWD`). Defaulting to `""` would break legitimate relative-path scripts without security gain. Happy to revisit in a follow-up if you'd like.

AlexandreYang (approved):

PR #233, opened 2026-05-06
test(awk): add rewritten oracle harness

matt-dz:
## Summary

Adds the rewritten AWK scenario harness and pinned GNU awk 5.4.0 oracle that future rshell awk implementation PRs build on. The harness keeps implementation gating opt-in through `enabled.txt`, runs rshell through an awk-shaped adapter, and preserves upstream provenance without requiring live upstream test fetches.

- Adds local rewritten AWK scenarios plus an upstream mapping ledger.
- Adds `tools/awk-harness` with pinned GNU awk install support and the `rshell-awk` adapter.
- Wires Make/CI entry points for metadata checks and rewritten scenario runs.
- Adds the `implement-awk` skill and compatibility guidance for future work.

## Validation

- `bash -n tools/awk-harness/*.sh tools/awk-harness/rshell-awk`
- `go test ./tests -run TestAwkScenarioMetadata -count=1`
- `make test_awk_rewritten`


.claude/skills/implement-awk/SKILL.md
```diff
@@ -0,0 +1,155 @@
+---
```
AlexandreYang, line 1: Topic from slack:

I'm wondering if could just compare to mawk already available in debian:bookworm-slim that we already use:
```
docker run --rm debian:bookworm-slim bash -c 'awk -W version'
random-funcs:       srandom/random
regex-funcs:        internal
compiled limits:
sprintf buffer      8192
maximum-integer     2147483647
mawk 1.3.4 20200120
Copyright 2008-2019,2020, Thomas E. Dickey
Copyright 1991-1996,2014, Michael D. Brennan
```

AlexandreYang: also:
I think it would be nice to make it work for scenarios tests "bash" comparison mode
which means possible installing gawk in debian:bookworm-slim or another image

.claude/skills/implement-awk/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+name: implement-awk
```
AlexandreYang, line 2: Topic 2 from slack:

I'm not fully sure if we should just use the upstream tests suites as it,

we might have/want behaviour that slight differ from upstream awk
features that we don't implement 
features we differ on purpose

---

also, beside the previous point, I believe, it's not ideal I believe to have this kind of dependency on upstream repo:
- slower CI
- less predictability, more flaky
- more complex setup

the good thing about replicating the tests as our own scenario tests is that:
- we have full control for divergent behaviour
- we can easily add more tests (if we use upstream, we can add our own scenarios tests too, but it's harder to track wasn't missing or not, and have 2 sources of tests)

matt-dz: Yeah, agreed that we shouldn’t run the upstream suites as-is or make them the main CI contract.

We do have harness filters, though, so I think the intended use is a curated allowlist/smoke profile: run only known-relevant upstream cases for the rshell awk subset, then translate any important discoveries into first-class tests/scenarios/ coverage.

So the split I’d suggest is:
- `tests/scenarios/` stay authoritative for normal CI and intentional rshell behavior.
- The upstream harness runs filtered cases only, either opt-in, periodic, or as a small curated CI job.
- We do not depend on the moving/full upstream suite for merge correctness.
- Any upstream failure we decide matters becomes an original rshell scenario test.

That gives us the benefit of gawk coverage without the full cost/flakiness/control problem

PR #237, opened 2026-05-07
feat(cd): add cd builtin with -L/-P symlink modes

julesmcrt:
## Summary

- Adds the `cd` builtin so rshell scripts can change directory inside `AllowedPaths`. Targets outside the sandbox are rejected; the working directory and `\$PWD`/`\$OLDPWD` are mutated atomically only after `sandbox.Stat` confirms the target is a directory.
- Supported flags: `-L` (default, lexical `..`), `-P` (physical, walks symlinks via the same sandbox-safe pattern as `pwd -P`), `-h`. Last-wins between `-L`/`-P` mirrors bash.
- Operands: bare `DIRECTORY`, `-` (changes to `\$OLDPWD` and prints the new path), no-arg (uses `\$HOME`). Multi-operand and empty operands are rejected. `-e`, `-@`, `CDPATH`, `~user`, and `cdable_vars` are intentionally unsupported.

The framework gains two `CallContext` callbacks — `ChangeDir(absDir) error` and `LookupEnvVar(name) (string, bool)` — the minimum surface needed for a builtin that mutates runner state, without exposing a generic env-write hook.

## Test plan

- [x] `go test ./...` (all green; new `builtins/cd/` package at 85% coverage)
- [x] Scenario tests in `tests/scenarios/cmd/cd/{basic,flags,errors,hardening,help}/`
- [x] Pentest probes: path traversal via `..`, escape via cross-root symlinks, unknown flags, flag injection via expansion, dash-prefixed filenames, FD-leak under 1000-iteration loops, stdin-ignored, no-filesystem-write
- [x] Fuzz seeds (`FuzzCdFlags`, `FuzzCdPath`) with corpora drawn from impl edge cases, CVE-class shapes, and existing test coverage; CI entry added to `.github/workflows/fuzz.yml`
- [x] `RSHELL_BASH_TEST=1` (Docker-backed bash compatibility) — could not run locally because the bind-mount under `t.TempDir()` is outside Docker Desktop's allow-list on this machine; please verify on CI
- [x] `make fmt` clean
- [x] `go run ./cmd/rshell --allow-all-commands -c 'help cd'` shows the new entry; `SHELL_FEATURES.md` updated alphabetically between `cat` and `continue`

## Notes

- Removed `tests/scenarios/shell/blocked_commands/builtins_and_features/cd.yaml` — that scenario asserted `cd` was an unknown command, which is no longer the case. A new `tests/scenarios/cmd/cd/errors/no_allowed_paths.yaml` covers the "cd with no sandbox" failure path.
- `cd -P` deliberately does **not** apply `CanonicalizeRootPrefix`. Unlike `pwd -P`, which uses canonicalisation for display, `cd` needs the working directory in the configured-root form so subsequent sandbox operations resolve under the same root. Documented in `resolvePath`.


julesmcrt:
@codex review

julesmcrt:
@codex review

interp/runner_exec.go
```diff
@@ -623,6 +623,8 @@ func (r *Runner) call(ctx context.Context, pos syntax.Pos, args []string) {
 				CommandAllowed: func(n string) bool {
 					return r.allowAllCommands || r.allowedCommands[n]
 				},
+				ChangeDir:    r.changeDir,
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in f80cc860 — the child CallContext used by RunCommand (find -exec, find -execdir, xargs) now leaves `ChangeDir` nil. cd in a sub-command exits 1 with "cd: not supported in this runner" rather than mutating the parent runner. Added `tests/scenarios/cmd/cd/hardening/cd_in_find_exec_isolated.yaml` to lock this in.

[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in f80cc860 — same root cause as #3201903210. The child CallContext used by RunCommand now leaves `ChangeDir` nil so cd in a sub-command cannot mutate the parent runner. New scenario `cd_in_find_exec_isolated.yaml` covers `find -execdir cd /elsewhere`.

interp/cd_support.go
```diff
@@ -0,0 +1,77 @@
...
+// unchanged and an error is returned.
+func (r *Runner) changeDir(absDir string) error {
+	if !filepath.IsAbs(absDir) {
+		return &os.PathError{Op: "chdir", Path: absDir, Err: errInvalidPath}
+	}
+	if r.sandbox == nil {
+		return &os.PathError{Op: "chdir", Path: absDir, Err: errNoSandbox}
+	}
+	cleaned := filepath.Clean(absDir)
+	info, err := r.sandbox.Stat(cleaned, r.Dir)
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in f80cc860 — `changeDir` now calls `sandbox.Access(cleaned, r.Dir, 0x01)` after the `IsDir()` check so a directory without search/execute permission is rejected with "permission denied", matching bash. `TestCdNoSearchPermissionRejected` in `builtins/cd/cd_unix_test.go` locks the behaviour in (Unix-only; root skipped because it bypasses permission bits).

julesmcrt:
@codex review

builtins/cd/cd.go
```diff
@@ -0,0 +1,459 @@
...
+
+	// -L and -P share a sequence counter so that after parsing we can
+	// compare their pos fields to determine which appeared last on the
+	// command line. pflag calls Set() in parse order, so the last flag
+	// Set gets the highest pos value. This mirrors pwd's pattern.
+	var modeSeq int
+	logicalFlag := newBoolSeqFlag(&modeSeq)
+	physicalFlag := newBoolSeqFlag(&modeSeq)
+	fs.VarPF(logicalFlag, "logical", "L", "treat .. lexically; preserve symbolic links in the result (default)").NoOptDefVal = boolSeqSentinel
+	fs.VarPF(physicalFlag, "physical", "P", "resolve symbolic links before treating ..").NoOptDefVal = boolSeqSentinel
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in f75e89b6 — `fs.SetInterspersed(false)` on the cd FlagSet so the first positional terminates flag parsing. `cd /tmp -P` now flows through the >1-positional check and exits 1 with "too many arguments", matching bash 5.2. Added `flag_after_directory_rejected.yaml` scenario, and dropped the skip_assert_against_bash on `too_many_args.yaml` since bash 5.2 also exits 1 (the original skip was based on bash 3.2 on macOS which exits 2).

julesmcrt:
@codex review

interp/cd_support.go
```diff
@@ -0,0 +1,92 @@
...
+	oldDir := r.Dir
+	r.Dir = cleaned
+	// Use setVarString so the new values count toward MaxVarBytes /
+	// MaxTotalVarsBytes the same way as a script-driven assignment.
+	// setVarString records storage-cap failures on the runner's exit
+	// state but does not return an error here; that's intentional —
+	// matching bash, the dir change is committed even if the env-var
+	// update could not be persisted.
+	r.setVarString("OLDPWD", oldDir)
+	r.setVarString("PWD", cleaned)
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in d0c288ec — two related corrections:

1. `cd_support.go` now reads `$PWD` (env var) as the source for `OLDPWD`, falling back to `r.Dir` only when unset. This matches bash, which captures the just-assigned `/bogus` from the temp `PWD=/bogus` rather than the real pre-cd dir.
2. `runner_exec.go` now skips the temp-assignment revert for `PWD`/`OLDPWD` when the dispatched command was a *successful* cd. A failed cd still reverts normally so the temp assignment cannot leak.

After both fixes:
- `cd /tmp/parent; PWD=/bogus cd b` → `PWD=/tmp/parent/b OLDPWD=/bogus` (verified against `debian:bookworm-slim` bash 5.2)
- `cd /tmp/parent; PWD=/bogus cd nope` → `PWD` reverts to `/tmp/parent` (matches bash)

Two new scenarios `temp_pwd_assignment.yaml` and `temp_pwd_failed_cd_reverts.yaml` lock both behaviours in and run against the bash compat job.

julesmcrt:
@codex review

builtins/cd/cd.go
```diff
@@ -0,0 +1,467 @@
...
+// resolveOperand inspects the parsed positional arguments and returns the
+// target path string. printResult is true when the caller must print the
+// resolved path on success (the `cd -` form). An error is returned when
+// the required env var (HOME / OLDPWD) is unset.
+func resolveOperand(args []string, callCtx *builtins.CallContext) (target string, printResult bool, err error) {
+	if len(args) == 0 {
+		// No-arg form: change to $HOME.
+		home, ok := lookupEnv(callCtx, "HOME")
+		if !ok || home == "" {
+			return "", false, errors.New("HOME not set")
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in 5f714a96 — `resolveOperand` now only errors when HOME is *unset*, not when it is set to the empty string. The handler short-circuits on an empty resolved target so `HOME= cd` is a no-op success (pwd unchanged), matching bash 5.2. New scenario `basic/home_empty_is_noop.yaml` runs against the bash compat job.

builtins/cd/cd.go
```diff
@@ -0,0 +1,467 @@
...
+
+	operand := args[0]
+	if operand == "" {
+		// Empty operand is rejected by POSIX.
+		return "", false, errors.New("invalid empty operand")
+	}
+	if operand == "-" {
+		oldpwd, ok := lookupEnv(callCtx, "OLDPWD")
+		if !ok || oldpwd == "" {
+			return "", false, errors.New("OLDPWD not set")
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in 5f714a96 — same root cause as #3202896970. `resolveOperand` now only errors when OLDPWD is unset, not when empty. `cd -` with `OLDPWD=` is a no-op success that prints a blank line, matching bash 5.2. New scenario `basic/oldpwd_empty_dash_prints_blank.yaml` runs against the bash compat job.

julesmcrt:
@codex review

builtins/cd/cd.go
```diff
@@ -0,0 +1,479 @@
...
+// resolved path. cd needs the working directory in the configured-root
+// form (e.g. /tmp/link) so subsequent sandbox operations resolve under
+// the same root; canonicalising would translate the path to the
+// host-canonical form (e.g. /var/data) which the sandbox would reject
+// as outside its configured roots. The trade-off is that cd -P cannot
+// resolve a symlink at the AllowedPaths root itself, only symlinks at
+// non-root components — a documented limitation.
+func resolvePath(ctx context.Context, callCtx *builtins.CallContext, absPath string, physical bool) (string, error) {
+	if !physical {
+		return filepath.Clean(absPath), nil
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in e4027926 — rewrote `resolvePath` to walk path components left-to-right and `StatFile` the accumulated prefix on every `..`. `cd no-such/..` and `cd file/..` now both exit 1 with bash-matching errors (verified against `debian:bookworm-slim`). The same walker handles `-L` and `-P`; `-P` additionally Lstats per-append and treats ErrPermission as opaque (above-sandbox passthrough) while ErrNotExist becomes a hard error. Also switched the relative-path join in the handler from `filepath.Join` (which Cleans) to manual concatenation so `..` survives into the walker. Two new bash-compared scenarios `errors/dotdot_after_missing.yaml` and `errors/dotdot_after_file.yaml`.

builtins/cd/cd.go
```diff
@@ -0,0 +1,479 @@
...
+
+		// bash 5.2 treats an empty target as a no-op success: cd ""
+		// stays in place, HOME='' cd stays in place, OLDPWD='' cd -
+		// stays in place but still prints an empty line. We mirror
+		// that contract here without consulting the filesystem.
+		if target == "" {
+			if printResult {
+				callCtx.Out("\n")
+			}
+			return builtins.Result{}
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in e4027926 — the empty-target branch no longer short-circuits before `ChangeDir`. Instead it rewrites the absolute target to cwd and lets ChangeDir refresh `$PWD`/`$OLDPWD` normally; the `cd -` form still prints just a newline. After `cd child; cd ""; cd -`, rshell now stays in `child` (matches bash 5.2). New scenario `basic/empty_cd_refreshes_oldpwd.yaml` runs against the bash compat job.

julesmcrt:
@codex review

interp/cd_support.go
```diff
@@ -0,0 +1,102 @@
...
+
+	r.Dir = cleaned
+	// Use setVarString so the new values count toward MaxVarBytes /
+	// MaxTotalVarsBytes the same way as a script-driven assignment.
+	// setVarString records storage-cap failures on the runner's exit
+	// state but does not return an error here; that's intentional —
+	// matching bash, the dir change is committed even if the env-var
+	// update could not be persisted.
+	r.setVarString("OLDPWD", oldDir)
+	r.setVarString("PWD", cleaned)
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in 907a191c — switched `changeDir` from `setVarString` (which stashes failures on r.exit.code that the cd builtin 's Result clobbers) to `setVarErr`, returning the error so cd surfaces exit 1. `TestCdReportsVarStorageCapFailure` pre-fills the env to `MaxTotalVarsBytes` and confirms `cd /path && echo CD_EXIT=$?` prints `CD_EXIT=1` together with the `variable storage limit exceeded` stderr.

julesmcrt:
@codex review

interp/cd_support.go
```diff
@@ -0,0 +1,108 @@
...
+	// differ when a temporary assignment precedes cd
+	// (e.g. `PWD=/bogus cd b`) — bash captures /bogus as OLDPWD even
+	// though the actual cwd was /tmp/parent. Fall back to r.Dir only
+	// when $PWD is unset.
+	oldDir := r.Dir
+	if v := r.writeEnv.Get("PWD"); v.IsSet() && v.Str != "" {
+		oldDir = v.Str
+	}
+
+	r.Dir = cleaned
```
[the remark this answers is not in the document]
julesmcrt: [Claude Opus 4.7 (1M context)] Fixed in d524f61b — reordered `changeDir` so the env writes are attempted *before* `r.Dir = cleaned`. If `setVarErr("OLDPWD", ...)` fails we return immediately with nothing mutated; if `setVarErr("PWD", ...)` fails after OLDPWD succeeded we restore OLDPWD via `setVarRestore` (which bypasses the cap) and then return — `r.Dir` is only committed once both env writes have succeeded. Matches bash: a failed cd leaves pwd untouched. `TestCdReportsVarStorageCapFailure` now asserts `POST_PWD=$dir` after the failed cd so a regression would surface immediately.

julesmcrt:
@codex review

builtins/cd/cd.go
```diff
@@ -0,0 +1,536 @@
...
+			// /private, /var, /folders, ... before re-entering the
+			// sandbox). Treat them as opaque pass-throughs so cd -P
+			// to a sandbox-internal target still resolves. Missing
+			// in-sandbox components surface as ErrNotExist instead
+			// and bubble up as a hard error, matching bash.
+			if errors.Is(err, iofs.ErrPermission) {
+				out = candidate
+				continue
+			}
+			return "", err
```
AlexandreYang, line 407: <sub><sub>P2 Badge</sub></sub> **`-P` does not resolve symlinks in above-sandbox path components — promote to package docstring**

The local comment explains why `iofs.ErrPermission` falls through to `out = candidate`, but the implication is subtle: `cd -P` only resolves symlinks whose *parent* is inside `AllowedPaths`. If a sandbox-internal target sits behind an above-sandbox symlink (e.g. on macOS where `/tmp` is itself a symlink to `/private/tmp`), `-P` will not resolve that hop — it silently behaves like `-L` for that segment.

This is the correct trade-off given the sandbox model, but it's worth surfacing in the package docstring (around `// Note: unlike pwd -P, cd does not apply CanonicalizeRootPrefix...`) so callers can reason about when `cd -P` diverges from bash on container-style sandboxes.

Not a security issue — final `changeDir` validates the result is in the sandbox — but bash-compat users may be surprised.

builtins/cd/cd.go
```diff
@@ -0,0 +1,536 @@
...
+			cleanedTarget := filepath.Clean(target)
+			// Container-style sandboxes mount the host filesystem at
+			// a prefix (e.g. /mnt/host). Symlink targets stored on
+			// disk often refer to host-absolute paths without that
+			// prefix, so apply the prefix when set.
+			if callCtx.HostPrefix != nil {
+				if hp := callCtx.HostPrefix(); hp != "" && !strings.HasPrefix(cleanedTarget, hp+string(filepath.Separator)) && cleanedTarget != hp {
+					cleanedTarget = filepath.Join(hp, cleanedTarget)
+				}
+			}
```
AlexandreYang, line 435: <sub><sub>P2 Badge</sub></sub> **`HostPrefix` re-prefix logic — consider an explicit hardening test**

The check `!strings.HasPrefix(cleanedTarget, hp+string(filepath.Separator)) && cleanedTarget != hp` is correct for the common case (host-absolute symlink target → prefix it), but I'd like an explicit test for:

- `hp=/mnt/host`, symlink target `/mnt/hostile/foo` → does this re-prefix to `/mnt/host/mnt/hostile/foo`? It does (prefix `/mnt/host/` doesn't match `/mnt/hostile/`).
- `hp=/mnt/host`, target exactly `/mnt/host` → the `!= hp` guard prevents re-prefix. ✓
- `hp=/mnt/host`, target `/mnt/host/foo` → `HasPrefix(/mnt/host/foo, /mnt/host/)` matches, no re-prefix. ✓

The `/mnt/hostile` collision is benign — sandbox.Stat will reject if the result is outside — but a hardening test asserting the documented behavior would lock the contract.

No scenario test currently exercises a non-empty `HostPrefix` with `cd -P` (scenarios run without `HostPrefix`); could be added at the Go-test level since scenario YAML doesn't expose the option.

builtins/cd/cd.go
```diff
@@ -0,0 +1,536 @@
...
+	return func(ctx context.Context, callCtx *builtins.CallContext, args []string) builtins.Result {
+		if *helpFlag {
+			printHelp(callCtx)
+			return builtins.Result{}
+		}
+
+		if callCtx.ChangeDir == nil || callCtx.WorkDir == nil {
+			callCtx.Errf("cd: not supported in this runner\n")
+			return builtins.Result{Code: 1}
+		}
```
AlexandreYang, line 166: <sub><sub>P3 Badge</sub></sub> **Improve the "not supported" diagnostic for the find -exec / xargs case**

The message `cd: not supported in this runner` is functionally correct but opaque from a user's perspective: it surfaces in `find -execdir cd ...` and `xargs cd ...` where `ChangeDir` is intentionally nil for isolation, and the user has no way to know why their `cd` failed.

Consider something like:
```
cd: cannot run inside `find -exec` / `xargs` — child invocations are isolated
```

Minor wording polish; behavior is fine.

tests/scenarios/cmd/cd/errors/home_unset.yaml
```diff
@@ -0,0 +1,12 @@
...
+  allowed_paths: ["$DIR"]
+  script: |+
+    unset HOME
+    cd
+expect:
+  stdout: ""
+  # bash: "bash: line 2: cd: HOME not set"
+  # ours: "cd: HOME not set"
+  stderr_contains: ["cd:", "HOME not set"]
+  exit_code: 1
```
AlexandreYang, line 12: <sub><sub>P2 Badge</sub></sub> **Coverage gap — `HOME` set but pointing outside the sandbox**

The current error scenarios cover `HOME` unset (this file) and `HOME=""` empty (`basic/home_empty_is_noop.yaml`). I don't see a scenario for `HOME=/outside-sandbox cd` — i.e. `HOME` set to a valid-looking path that the sandbox rejects.

Suggested addition (e.g. `errors/home_outside_sandbox.yaml`):
```yaml
description: cd with no operand uses $HOME; an out-of-sandbox HOME is rejected with permission denied.
skip_assert_against_bash: true  # bash has no sandbox
input:
  allowed_paths: ["$DIR"]
  script: |+
    HOME=/etc cd
expect:
  stdout: ""
  stderr_contains: ["cd:", "permission denied"]
  exit_code: 1
```

This exercises the path where `resolveOperand` succeeds but `changeDir` rejects — distinct from the unset-HOME path.

interp/cd_support.go
```diff
@@ -0,0 +1,115 @@
...
+	//
+	// Skipped on Windows: there are no POSIX execute bits, and
+	// allowedpaths.accessCheck always denies an execute check on Windows.
+	// Directory traversal is governed by ACLs that os.Root has already
+	// honoured at sandbox open, so the Stat above is a sufficient guard.
+	if runtime.GOOS != "windows" {
+		if err := r.sandbox.Access(cleaned, r.Dir, 0x01); err != nil {
+			return err
+		}
+	}
```
AlexandreYang, line 69: <sub><sub>P3 Badge</sub></sub> **Windows access-check skip — consider an explicit Windows test**

The rationale (`os.Root` honours ACLs at open time, `accessCheck` always denies execute on Windows) is sound. The Windows test file (`cd_windows_test.go`) covers reserved-name rejection, NTFS case-insensitivity, and forward-slash paths, but I didn't see a test that exercises a sandbox-rooted directory the current user lacks traversal rights on.

Not blocking — Windows ACL test setup is fiddly. Just flagging that the documented invariant (`os.Root` ACL enforcement at open) isn't directly asserted here. The Unix counterpart is `TestCdNoSearchPermissionRejected`.

interp/runner_exec.go
```diff
@@ -116,7 +116,19 @@ func (r *Runner) cmd(ctx context.Context, cm syntax.Command) {
...
 		defer func() {
+			// cd intentionally writes $PWD and $OLDPWD as part of
+			// its semantics. Reverting those after a successful cd
+			// would leave the env vars disagreeing with the shell's
+			// tracked working directory — bash skips the revert in
+			// the same case (e.g. `PWD=/bogus cd b` keeps PWD at
+			// the new dir afterwards). The skip is scoped to a
+			// successful cd so a cd that errored still gets its
+			// temp PWD assignment reverted normally.
+			isCd := len(fields) > 0 && fields[0] == "cd" && r.exit.ok()
```
AlexandreYang, line 127: <sub><sub>P3 Badge</sub></sub> **`isCd` defer skip is keyed on the literal command name**

The check `fields[0] == "cd"` is fine today because rshell has no aliases, no `command cd ...` indirection, and no `function`-keyword name override. If any of those are added later, the defer-skip would silently regress (a successful `cd` reached via the new path would revert the env writes it just made).

Not actionable today — just flagging it as a low-priority invariant to test if the feature surface ever grows.

AlexandreYang (commented):
## Code Review — `cd` builtin (PR #237)

**Scope:** New `cd` builtin (`builtins/cd/`), framework callbacks (`ChangeDir`, `LookupEnvVar`), runner integration in `interp/cd_support.go` and `interp/runner_exec.go`, plus 58 scenario tests, 11 hardening scenarios, fuzz seeds, and Go unit / pentest / bash-compat tests.

**Overall:** Carefully engineered. The implementation walks paths component-by-component through the sandbox's `Stat`/`Lstat`/`Readlink` callbacks (no direct `os.*` filesystem calls), respects bash semantics for `-L`/`-P` last-wins flag ordering and `cd ""` no-op refresh, applies env writes before committing `r.Dir`, isolates `cd` in `RunCommand` children (find -exec, xargs), skips the execute-bit check on Windows where `os.Root` ACL checks cover it, and bounds symlink expansion at 40 hops. Test coverage is extensive — 58 scenarios with most asserting against bash, plus pentest + fuzz + bash-compat Go tests.

No P0/P1 findings. A small number of P2/P3 observations follow as inline comments — primarily around documentation/test surface area for subtle sandbox-spanning interactions.

### Summary of findings

| 1 | <sub><sub>P2 Badge</sub></sub> | `builtins/cd/cd.go:384-388` | `-P` mode does not resolve above-sandbox symlinks (Lstat ErrPermission → opaque pass-through). Documented locally; worth promoting to the package docstring as a sandbox-spanning invariant. |
| 2 | <sub><sub>P2 Badge</sub></sub> | `builtins/cd/cd.go:412-416` | `HostPrefix` re-prefix check matches `hp+sep` — a target whose canonical path collides with a sibling of the host prefix can get re-prefixed. Worth a hardening test if `HostPrefix` is in scope. |
| 3 | <sub><sub>P2 Badge</sub></sub> | `tests/scenarios/cmd/cd/errors/` | Coverage gap — no scenario asserts that `HOME=/outside-sandbox cd` is rejected with permission denied. Current tests only cover HOME unset and HOME empty. |
| 4 | <sub><sub>P3 Badge</sub></sub> | `builtins/cd/cd.go:144-147` | The `not supported in this runner` message is correct but technical — under `find -execdir cd ...` users will see this when the cause is intentional isolation. Consider phrasing that hints at the cause. |
| 5 | <sub><sub>P3 Badge</sub></sub> | `interp/cd_support.go:65-69` | Windows access-check skip is correctly documented (`os.Root` handles ACLs). Optional: a Windows-specific test that exercises a sandbox-rooted denied directory would close the loop. |

### Positive observations

- All filesystem reads in `resolvePath` go through `callCtx.StatFile` / `LstatFile` / `ReadlinkFile`. No direct `os.*` filesystem calls in builtin code.
- `changeDir` orders env writes *before* committing `r.Dir`, with explicit `OLDPWD` rollback on `PWD` write failure — verified by the dedicated `TestCdReportsVarStorageCapFailure` Go test against the `MaxTotalVarsBytes` cap.
- Per-component `.. ` validation (line 348) correctly rejects `cd file/..` and `cd no-such/..`, matching bash 5.2 — both directly tested.
- `boolSeqFlag` uses a private sentinel for `NoOptDefVal` so `cd --physical=true` is rejected, matching bash.
- `RunCommand` child contexts (find -exec, xargs) set `ChangeDir: nil` to isolate `cd` from leaking into the parent shell — verified by `cd_in_find_exec_isolated.yaml`.
- The pipeline subshell case (in `runner_exec.go`) uses `subshell(true)` so cd inside `(...)` or a pipeline stage mutates only the child's `r.Dir`.
- Fuzz seeds cover flag injection, control-character rejection (via `shellSafe`), long inputs, CVE-class shapes; both `FuzzCdFlags` and `FuzzCdPath` assert no panics and only valid exit codes.

builtins/cd/cd.go
julesmcrt: [Claude Opus 4.7 (1M context)] Done in 1f1f731d — promoted the `-P` above-sandbox passthrough invariant into the package docstring, including the macOS `/tmp` → `/private/tmp` example. Sandbox-rooted callers reading `cd.go` now see the divergence-from-bash trade-off upfront rather than buried inside `resolvePath`.

builtins/cd/cd.go
julesmcrt: [Claude Opus 4.7 (1M context)] Done in 1f1f731d — added four `TestHostPrefix*` unit tests in `cd_internal_test.go` exercising the re-prefix branch with a stubbed CallContext:

- `…RePrefixesUnrelatedAbsoluteTarget` — `/var/log/...` → `/mnt/host/var/log/...`
- `…SkipsRePrefixWhenAlreadyPrefixed` — `/mnt/host/var/...` left alone
- `…SkipsRePrefixWhenExactlyTheHostPrefix` — target equal to `hp` left alone
- `…RePrefixesPathPrefixCollision` — the `/mnt/host` vs `/mnt/hostile` boundary you specifically flagged: with `hp+sep` the latter does get re-prefixed (correct, the final sandbox.Stat then rejects). All four pass.

builtins/cd/cd.go
julesmcrt: [Claude Opus 4.7 (1M context)] Done in 1f1f731d — changed the diagnostic from `cd: not supported in this runner` to `cd: cannot change directory from inside find -exec/-execdir or xargs (child invocations are isolated)`. Users hitting this through `find -execdir cd ...` or `xargs cd ...` now see the cause and the intentional-isolation rationale.

tests/scenarios/cmd/cd/errors/home_unset.yaml
julesmcrt: [Claude Opus 4.7 (1M context)] Done in 1f1f731d — added `tests/scenarios/cmd/cd/errors/home_outside_sandbox.yaml` exercising `HOME=/etc cd` with `allowed_paths: ["$DIR"]`: stderr contains `cd:` + `permission denied`, exit 1. Documented in the description as the scenario that closes the gap between the unset-HOME path and the empty-HOME no-op path. Marked `skip_assert_against_bash: true` since bash has no sandbox.

interp/cd_support.go
julesmcrt: [Claude Opus 4.7 (1M context)] Acknowledged — leaving this open. As you noted, the invariant is enforced by `os.Root`'s ACL handling at open time so the on-the-wire contract is correct, but a Windows-specific regression test would close the loop. Holding off because (a) reliable Windows ACL test setup needs DENY ACEs that GH-actions runners may not allow under the default user, and (b) the macOS/Linux `TestCdNoSearchPermissionRejected` covers the Unix counterpart of the same contract. Happy to add it if you have a pattern that worked elsewhere in the repo.

interp/runner_exec.go
julesmcrt: [Claude Opus 4.7 (1M context)] Agreed — this is forward-looking and not actionable today, leaving the thread open. The `fields[0] == "cd"` literal check would silently regress if rshell ever grew aliases, `command cd`, or shell-function name override. Worth a `// TODO` comment near that defer if/when one of those surfaces; for now flagged so future contributors know the invariant.

AlexandreYang:
@codex make a comprehensive code and security review

AlexandreYang (approved):

PR #242, opened 2026-05-07
fix(help): list while/until as supported in help output

AlexandreYang:
## Summary

- The `control-flow` feature in `help` advertised `while` and `until` as unsupported, but the interpreter actually supports both (and `SHELL_FEATURES.md` already documents them as ✅).
- Verified by running scripts like `X=0; while [ "$X" != stop ]; do echo iter; X=stop; done` against the built binary — they iterate normally and exit 0.
- Moved both forms from the Unsupported list to the Supported list under `control-flow`, refreshed the feature one-liner (`for/while/until/if/...; no case/select/functions.`), and dropped `while/until` from the top-level unsupported summary.
- Updated the pinned strings in `builtins/tests/help/help_test.go` and `tests/scenarios/cmd/help/unrestricted.yaml` so they assert the new wording.

All other commands and features listed in `help` were verified to be present (33 commands, 8 features) — this was the only mismatch.

## Test plan

- [x] `go test ./builtins/... ./tests/...` (passes locally)
- [x] `make fmt` clean
- [x] `rshell -c 'help control-flow'` shows `while`/`until` under Supported
- [x] `rshell -c 'help'` summary no longer claims `while/until` is unsupported


AlexandreYang:
@codex review

astuyve (approved):

astuyve:
codex is gonna say it's good

PR #244, opened 2026-05-08
feat(awk): implement phase 1 core interpreter

matt-dz:
## Summary

Implements the first practical awk slice on top of the rewritten harness, adding the registered `awk` builtin with core parsing and evaluation. The phase covers BEGIN/END, record and field reads, simple separators, regex/comparison patterns, arithmetic and string behavior, and the GNU awk oracle wiring needed to keep the implementation honest.

- Adds the `builtins/awk` implementation and registers the command.
- Supports core printing, fields, patterns, assignments, arithmetic, and string concatenation.
- Enables the first 107 rewritten AWK scenarios plus GNU awk oracle shell scenarios.
- Adds pinned GNU awk CI jobs and awk-focused fuzzing.

## Validation

- `make test_against_gawk`
- `make test_awk_rewritten`
- `tools/awk-harness/run.sh fuzz`


matt-dz:
Superseded by #599, which consolidates the full AWK implementation on current main.

PR #524, opened 2026-06-09
feat(interp): add RemediationMode() RunnerOption [PR A]

julesmcrt:
## Summary

- Adds `RemediationMode() RunnerOption` in `interp/api.go` with the corresponding `remediationMode bool` field on `runnerConfig`
- Adds `--remediation-mode` boolean CLI flag (default `false`) in `cmd/rshell/main.go`
- Documents the new mode in `README.md` and `SHELL_FEATURES.md` (one line each)

**This PR is purely an API addition — no behavior changes.** The option is accepted and stored on the runner but is otherwise inert.

## Context (multi-PR series)

| PR | Description |
|----|-------------|
| **A (this PR)** | `RemediationMode()` option — API only, inert |
| B | Add write capability to `allowedpaths/sandbox.go` |
| C | Wire `remediationMode` into interpreter so `>` / `>>` redirects work when set |
| Later | `truncate` and `logrotate` builtins |

End goal: a `runRemediationCommand` PAR action creates a `Runner` with `RemediationMode()` set; the existing `runCommand` action does not.

## Test plan

- [x] `go build ./...` — clean
- [x] `go test ./...` — all pass
- [x] `make fmt` — no changes after formatting

PR #525, opened 2026-06-10
allowedpaths: relax sandbox to permit write opens (remediation series PR B)

julesmcrt:
## Summary

- Adds `allowedOpenFlags` const (`O_RDONLY|O_WRONLY|O_RDWR|O_APPEND|O_CREATE|O_EXCL|O_TRUNC`) to `allowedpaths/sandbox.go`
- Changes `Sandbox.Open` to reject unknown flag bits (instead of blanket non-`O_RDONLY` rejection), allowing write opens inside the allowlist
- Cross-root symlink fallback remains read-only to prevent TOCTOU attacks on writes
- New tests: write to allowed path, write outside allowed path, append, truncate, symlink-escape rejected for writes, unknown flag rejected, read still works
- Updates `analysis/symbols_allowedpaths.go` with the new OS flag constants
- Updates `README.md`, `SHELL_FEATURES.md`, and `interp/api.go` doc comment to reflect that both reads and writes are now sandboxed

## Context

This is **PR B** of the remediation mode series. PR A (#524) added the `RemediationMode()` option to `interp/api.go`.

The sandbox is now write-capable, but nothing in the interpreter uses it yet — output redirections (`>`, `>>`) remain blocked at parse time. Wiring write opens into shell features (redirections, builtins) is PR C.

## Test plan

- [x] `go test ./allowedpaths/... ./interp/... -timeout 120s` passes
- [x] New sandbox write tests cover: create, append, truncate, symlink-escape rejection, unknown flag rejection, read still works
- [x] No interpreter-level behavior changes — existing scenario tests unaffected


val06 (approved):

allowedpaths/sandbox.go
```diff
@@ -334,11 +334,29 @@ func IsDevNull(path string) bool {
...
+// allowedOpenFlags are accepted; anything else returns ErrPermission.
+//
+// The cross-root symlink fallback (resolveFollowingSymlinks) is read-only.
+// Following a symlink that escapes its os.Root and then performing a write
+// (O_CREATE/O_TRUNC/O_APPEND/O_WRONLY/O_RDWR) is the classic TOCTOU footgun:
+// a malicious symlink could redirect a create or truncate to a target that
+// has changed between resolution and open, defeating the sandbox. Writes
+// must therefore stay within a single os.Root and never traverse the
+// cross-root fallback.
 func (s *Sandbox) Open(path string, cwd string, flag int, perm os.FileMode) (io.ReadWriteCloser, error) {
```
AlexandreYang, line 367: Can/Shall we have a mechanism to have an allowed list based on shell features?

For example, an allowed commands that says `>` `>>` are allowed to use `OpenWithWrite()` for example.

Just the current change in the PR, it seems that, we don't have clear visibility which shell feature is allowed and will use open with write os.FileMode.

julesmcrt: The next PR should match what you have in mind ;) https://github.com/DataDog/rshell/pull/526

AlexandreYang (approved):

allowedpaths/sandbox.go
AlexandreYang, line 367: Can/Shall we also pass remediation mode state to the Sandbox and verify that if read-only mode, to block write os.FileMode flags?

allowedpaths/sandbox.go
```diff
@@ -334,11 +334,29 @@ func IsDevNull(path string) bool {
 	return false
 }
 
+// allowedOpenFlags is the set of os.OpenFile flag bits permitted by the
+// sandbox. Anything outside this mask (O_DIRECTORY, O_NOFOLLOW, O_SYNC,
+// O_NONBLOCK, etc.) is rejected as a defense-in-depth measure so unknown
+// or platform-specific flags cannot slip through.
+const allowedOpenFlags = os.O_RDONLY | os.O_WRONLY | os.O_RDWR |
+	os.O_APPEND | os.O_CREATE | os.O_EXCL | os.O_TRUNC
```
AlexandreYang: check to only keep minimal needed

julesmcrt: Removed `O_RDWR` and `O_EXCL`
allowedOpenFlags trimmed to `O_RDONLY | O_WRONLY | O_APPEND | O_CREATE | O_TRUNC` exactly what shell redirections need

A separate writeOpenFlags const (with `O_WRONLY | O_APPEND | O_CREATE | O_TRUNC`) exists

allowedpaths/sandbox.go
julesmcrt: Agreed, `Sandbox` now has a `readOnly` bool field (default `true`)

PR #536, opened 2026-06-22
feat(builtins): add logrotate

matt-dz:
## Summary

Adds a remediation-only `logrotate` builtin for rshell that keeps the familiar logrotate intent while staying explicit about its safe subset: it truncates log files to zero bytes through `AllowedPaths`, but does not parse configs, rename/retain copies, compress logs, write state files, or run rotate scripts.

Key behavior:
- supports `--size/-s`, `--force/-f`, `--dry-run/-n`, `--verbose/-v`, and `--help/-h`
- requires exactly one of `--size` or `--force`
- dry-run reports per-file actions without mutating files
- missing files are errors and are not created
- threshold checks and truncation happen on the same fd via a new narrow sandbox capability
- write targets that traverse symlinks are rejected with `symlinks are not supported as write targets`

Also extracts shared coreutils-style size parsing for `truncate` and `logrotate`, wires the new write-capable `CallContext` field through remediation mode only, and updates docs/analyzer metadata.

## Validation

- `make fmt`
- `go test ./allowedpaths ./interp`
- `go test ./tests/ -run TestShellScenarios -timeout 120s`
- `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash -timeout 120s`
- `go test ./...`


matt-dz:
@codex review

builtins/logrotate/logrotate.go
```diff
@@ -0,0 +1,175 @@
...
+
+		if failed {
+			return builtins.Result{Code: 1}
+		}
+		return builtins.Result{}
+	}
+}
+
+func dryRunFile(ctx context.Context, callCtx *builtins.CallContext, file string, threshold int64, report bool) bool {
+	info, err := callCtx.StatFile(ctx, file)
```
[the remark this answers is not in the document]
matt-dz: Fixed in `00cd10a6` by making `--dry-run` call the same `TruncateToZeroIfAtLeast` sandbox capability as real truncation, with a dry-run flag that performs the write-target open/fstat validation and eligibility check without calling `ftruncate`. This means dry-run now rejects read-only roots and other write-target failures instead of reporting a false success.

Regression coverage:
- `tests/scenarios/cmd/logrotate/dry_run/read_only_root.yaml`

Validation:
- `make fmt`
- `go test ./tests/ -run TestShellScenarios/cmd/logrotate -timeout 120s`
- `go test ./builtins/logrotate ./allowedpaths ./interp ./analysis -timeout 120s`
- `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash -timeout 120s`
- `go test ./... -timeout 120s`

matt-dz:
@codex review

analysis/symbols_builtins.go
```diff
@@ -379,12 +383,9 @@ var builtinPerCommandSymbols = map[string][]string{
 		"context.Context", // 🟢 deadline/cancellation plumbing; pure interface, no side effects.
 	},
 	"truncate": {
```
julesmcrt, line 385: Why are we changing the symbols used by `truncate` here?

analysis/symbols_builtins.go
```diff
@@ -556,15 +557,16 @@ var callCtxAllFields = []string{
...
 }
 
 // builtinPerCommandCallContextFields maps each builtin command name to the
 // CallContext function fields it is permitted to access.
 //
 // Security invariants enforced by this map:
-//   - "Truncate" (the only write-capable CallContext field) must appear only in
-//     the "truncate" entry.
+//   - Write-capable CallContext fields ("Truncate",
+//     "TruncateToZeroIfAtLeast") must appear only in remediation-only entries.
```
julesmcrt: Is this comment good? Shouldn't _not_ mention Truncate and just talk about write-capable builtins?

julesmcrt (changes requested):
Shouldn't we split the PR?
The `:ro/:rw` path-mode suffix feels like it should be a standalone change

julesmcrt:
From Claude review:

`allowedpaths/internal/writeopen/writeopen_unix.go` — Symlinked log files get a confusing ELOOP error

openWriteFile applies O_NOFOLLOW to the final path component, so any file that is a symlink (even within a :rw root) fails with ELOOP ("too many levels of symbolic links"). In logrotate's context this is particularly visible: logrotate --dry-run --force /var/log/app.log where app.log is a common → /data/logs/app.log-style managed symlink would print:

logrotate: "app.log": too many levels of symbolic links

rather than something actionable. This is intentional per the symlink policy, but the error surfaced to the user doesn't say "symlinks are not supported as write targets." Adding a specific check and error before the openWriteFile call would improve diagnosability. Also worth a note in the help text / SHELL_FEATURES.md entry.

allowedpaths/sandbox.go
```diff
@@ -123,9 +147,39 @@ func (s *Sandbox) resolve(absPath string) (*root, string, bool) {
...
-		return &s.roots[i], rel, true
+		if best == nil || len(s.roots[i].absPath) > len(best.absPath) {
+			best = &s.roots[i]
+			bestRel = rel
+		}
+	}
+	return best, bestRel, best != nil
+}
+
+func (s *Sandbox) resolveCanonical(absPath string) (*root, string, bool) {
```
julesmcrt, line 142: resolve and resolveCanonical are structurally identical, creating a maintenance hazard

Both functions loop over s.roots, compute filepath.Rel, do the same ..-escape check, and pick the longest-prefix winner. They differ only in which field is used (absPath vs canonicalAbsPath) and that resolveCanonical has one extra tie-breaking rule (equal-length canonical paths → prefer :ro). Any future bug fix in resolve (e.g. handling trailing-slash edge cases, or adding a depth cap) must be manually mirrored. A single generic helper parameterized by a field selector and an optional tie-breaker would eliminate the duplication.

builtins/logrotate/logrotate.go
```diff
@@ -0,0 +1,152 @@
...
+// Accepted flags:
+//
+//	-s SIZE, --size=SIZE
+//	    Only truncate files whose current size is at least SIZE bytes. SIZE
+//	    uses the same non-negative coreutils suffix grammar as truncate -s.
+//
+//	-f, --force
+//	    Truncate without a size threshold. Mutually exclusive with --size.
+//
+//	-n, --dry-run
```
julesmcrt, line 27: --force --dry-run on an already-empty file reports "would truncate 0 bytes"

TruncateToZeroIfAtLeast(minSize=0) returns (0, true, nil) for any file size ≥ 0, including empty files. With --force --dry-run on a 0-byte log, the output is:

logrotate: empty.log: would truncate 0 bytes

This falsely implies there's something to remove. The fix is a simple guard: if sizeBefore == 0 and the file is already empty, return (0, false, nil) — nothing to do. No test covers this edge case.

analysis/symbols_builtins.go
matt-dz: This is just a shared-parser extraction, not an intended behavior change to `truncate`. `truncate` already had the coreutils-style size parser that `logrotate --size` needs, so I moved that parser into `builtins/internal/sizeparse` and kept the local `truncate` `parseSize` wrapper delegating to it. The symbol allowlist changed only because `truncate` now imports the shared parser instead of carrying the parser implementation locally.

builtins/logrotate/logrotate.go
matt-dz: Fixed in `84224502`. Empty files now return a no-op from `TruncateToZeroIfAtLeast`, so `--force --dry-run` no longer reports `would truncate 0 bytes`. When reporting is enabled, logrotate prints `already empty, skipping`; threshold skips with a positive `--size` still use the below-threshold message.

Regression coverage:
- `tests/scenarios/cmd/logrotate/dry_run/empty_force.yaml`
- `tests/scenarios/cmd/logrotate/basic/empty_force_verbose.yaml`

Validation:
- `make fmt`
- `go test ./tests/ -run TestShellScenarios/cmd/logrotate -timeout 120s`
- `go test ./builtins/logrotate ./allowedpaths ./interp ./analysis -timeout 120s`
- `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash -timeout 120s`

allowedpaths/sandbox.go
matt-dz: this applies to the stacked PR so i'll address it there and not in this PR

matt-dz:
@codex review

allowedpaths/write_open.go
```diff
@@ -20,5 +21,19 @@ func closeWriteRoot(file *os.File) {
...
+	if err := r.rejectSymlinkWriteTarget(relPath); err != nil {
+		return nil, err
+	}
 	return writeopen.OpenFile(r.writeRoot, r.root, relPath, flag, perm)
 }
+
+func (r *root) rejectSymlinkWriteTarget(relPath string) error {
+	info, err := r.root.Lstat(relPath)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
matt-dz: Fixed in `fa5852059b0a5d509059108af3c57c0cf2d5e53c`. `rejectSymlinkWriteTarget` now walks each path component with `Lstat` before opening, so symlinked parent directories like `link/app.log` return `symlinks are not supported as write targets` instead of leaking the platform `openat` error.

Regression coverage:
- `TestSandboxSymlinkWriteDirectoryComponentReportsActionableError`
- `TestLogrotateSymlinkDirectoryComponentReportsActionableError`

Validation:
- `make fmt`
- `go test ./allowedpaths ./interp`
- `go test ./tests/ -run TestShellScenarios -timeout 120s`
- `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash -timeout 120s`
- `go test ./...`

matt-dz:
@codex review

julesmcrt (approved):

PR #557, opened 2026-07-17
feat: add vmstat builtin for host memory/CPU/IO pressure investigation

Hitsuji-M:
## Summary
- Add a `vmstat` builtin for host memory/CPU/IO pressure investigation, mirroring the existing `df`/`ss` architecture.
- Linux reads `/proc/{stat,meminfo,vmstat,loadavg,uptime}` directly (sandbox-exempt, same pattern as `ss`/`ip route`/`df`).
- macOS reads via `sysctl(3)` (`hw.memsize`, `vm.swapusage`, `vm.loadavg`) using the existing darwin toolset, not Mach `host_statistics64`; unavailable columns render as `0`/`-` rather than being fabricated.
- Windows/other platforms report "not supported".
- Supports three modes: default snapshot, `vmstat DELAY [COUNT]` sampling, and `-s`/`--stats` full counter summary, plus `-a`, `-w`, `-S k|K|m|M`, `-h`/`--help`.
- Updates `analysis/symbols_builtins.go` and `analysis/symbols_internal.go` allowlists, `SHELL_FEATURES.md`, `README.md`, and `AGENTS.md` security design decisions.

## Test plan
- [x] `make fmt` / `gofmt -l .` clean
- [x] `go test ./analysis/... ./builtins/... ./tests/...`
- [x] Black-box tests over synthetic `/proc` fixtures (`builtins/tests/vmstat/vmstat_linux_test.go`, run via Docker `golang:1.25`)
- [x] Scenario tests under `tests/scenarios/cmd/vmstat/` (all `skip_assert_against_bash: true` — vmstat isn't in `debian:bookworm-slim` and output is host-state-dependent)
- [x] Manual verification on macOS: default snapshot, `-a`, `-w`, `-s`, `--help`


builtins/vmstat/vmstat.go
```diff
@@ -0,0 +1,536 @@
...
+	d, convErr := strconv.ParseUint(args[0], 10, 32)
+	if convErr != nil || d == 0 {
+		return 0, 0, fmt.Errorf("invalid delay '%s'", args[0])
+	}
+	if len(args) == 2 {
+		c, convErr := strconv.ParseUint(args[1], 10, 32)
+		if convErr != nil || c == 0 {
+			return 0, 0, fmt.Errorf("invalid count '%s'", args[1])
+		}
+		count = int(c)
```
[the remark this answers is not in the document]
Hitsuji-M: Fixed in [49bbf62](https://github.com/DataDog/rshell/pull/557/commits/49bbf62fac250aa7ec120f94eee5d2e61ed8278e)

builtins/internal/vmstat/vmstat_other.go
```diff
@@ -0,0 +1,16 @@
...
+//go:build !linux && !darwin
+
+package vmstat
+
+import "context"
+
+// readImpl returns ErrNotSupported on platforms without a backend
+// (Windows and anything else). Mirrors diskstats_other.go.
+func readImpl(_ context.Context, _ string) (Stats, error) {
+	return Stats{}, ErrNotSupported
```
matt-dz, line 15: what's the reason for not supporting windows? asking because i'm unfamiliar with the platform and want to understand the decision

matt-dz (commented):
looks good for the most part. however, i recommend running a review loop with codex -- it catches more in-the-weeds issues with the builtin logic

builtins/internal/vmstat/vmstat_other.go
Hitsuji-M: vmstat does not exist for Windows

san-jos:
@codex review

matt-dz:
@codex review

builtins/vmstat/vmstat.go
```diff
@@ -0,0 +1,571 @@
...
+	return fmt.Sprintf("%*d", width, v/uint64(divisor))
+}
+
+func dash(width int) string {
+	return fmt.Sprintf("%*s", width, "-")
+}
+
+// subClamp returns a-b, clamped to 0 instead of underflowing when b > a
+// (e.g. an adversarial or momentarily-inconsistent kernel counter read).
+func subClamp(a, b uint64) uint64 {
```
julesmcrt, line 515: <sub><sub>P2 Badge</sub></sub> **Missing direct unit tests for arithmetic helpers**

`subClamp`, `satUint64`, `cpuPercents`, `pct`, and `rateSwap`/`rateIO`/`rateSystem` implement exactly the integer-safety guarantees `docs/RULES.md` asks for (overflow saturation, counter-reset clamping, tick-percentage rounding), but they're currently only exercised indirectly through integration tests built on one fixed synthetic `/proc` snapshot — which never triggers a counter rollover, a `prev > cur` reset, or a saturating overflow.

Suggest a `builtins/vmstat/vmstat_test.go` with table tests, e.g.:
- `subClamp(5, 10) == 0` (counter reset / adversarial read)
- `satUint64` around `math.MaxUint64` boundaries
- `cpuPercents` with `prev` counters larger than `cur` (simulated counter reset)
- `pct` rounding edge cases where `us+sy+wa+st` would exceed 100 before the `id` clamp

builtins/vmstat/vmstat.go
```diff
@@ -0,0 +1,571 @@
...
+	}
+
+	prev := first
+	for i := 1; count == 0 || i < count; i++ {
+		timer := time.NewTimer(delay)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return builtins.Result{Code: 1}
+		case <-timer.C:
```
julesmcrt, line 269: <sub><sub>P3 Badge</sub></sub> **Unbounded sampling loop's context-cancellation path is untested**

Every existing sampling scenario/test passes an explicit `count`, so the `count == 0` (unbounded, "run until interrupted") branch of this loop — the one DoS-relevant path unique to this builtin — never actually gets exercised against a cancelled/timed-out `context.Context`. A quick Go test with `context.WithTimeout` and `vmstat 1` (no count) would confirm the loop actually stops and returns promptly rather than relying only on manual verification.

builtins/vmstat/vmstat.go
```diff
@@ -0,0 +1,571 @@
...
+
+		if *f.stats {
+			return runStats(ctx, callCtx, divisor, *f.unit)
+		}
+		return runSampling(ctx, callCtx, *f.active, *f.wide, divisor, delay, count)
+	}
+}
+
+// unitDivisor maps the -S argument to a byte divisor.
+func unitDivisor(s string) (int64, error) {
```
julesmcrt, line 165: <sub><sub>P3 Badge</sub></sub> **`-S k`/`-S K` branches aren't distinguished in tests**

`tests/scenarios/cmd/vmstat/flags/unit_scale.yaml` only compares `-S M` against `--unit=m`. The `k` (1000) and `K` (1024, also the default) cases of `unitDivisor` have no scenario asserting their scaled output actually differs from one another (analogous to how the `m`/`M` case is checked).

julesmcrt (commented):
Reviewed the `vmstat` builtin end to end (internal Linux/macOS/other backends, the command layer, docs, and tests).

**Security**: No sandbox-integrity issues. The Linux backend's direct `os.Open` on `/proc/{stat,meminfo,vmstat,loadavg,uptime}` and the macOS `sysctl(3)` reads are hardcoded, non-user-controllable paths — consistent with the documented `df`/`ss`/`ip route`/`free` exception, and `AGENTS.md`/`README.md` were updated accordingly. Integer-safety guards are solid: `subClamp` prevents underflow on counter resets, `satUint64` saturates instead of relying on undefined float-to-uint64 conversion, and the meminfo KiB→bytes conversion is bounds-checked against overflow before multiplying. The unbounded `vmstat <delay>` sampling loop correctly selects on `ctx.Done()` each iteration, so it's bounded by the shell's execution timeout rather than looping forever.

**Bash compat**: N/A — `vmstat` isn't present in `debian:bookworm-slim`, so all host-state-dependent scenarios correctly set `skip_assert_against_bash: true`.

**Correctness**: The `Partial` bitmask design (gating field groups independently, e.g. `FieldMemory` vs `FieldMemoryDetail`) correctly avoids the false-zero trap that `free`'s AGENTS.md entry calls out for macOS. The darwin `xsw_usage`/`loadavg` struct byte-layouts and offsets check out against the actual C struct definitions.

A few non-blocking test-coverage gaps below — nothing that should block merge.

| 1 | <sub><sub>P2 Badge</sub></sub> | `builtins/vmstat/vmstat.go:461` | Arithmetic helpers (`subClamp`, `satUint64`, `cpuPercents`, `pct`, rate functions) have no direct unit tests |
| 2 | <sub><sub>P3 Badge</sub></sub> | `builtins/vmstat/vmstat.go:218` | Unbounded sampling (`vmstat 1`, no count) isn't tested for context-cancellation cutoff |
| 3 | <sub><sub>P3 Badge</sub></sub> | `builtins/vmstat/vmstat.go:155` | `-S k`/`-S K` scaling branches aren't distinguished from each other in scenario tests |

Overall: **safe to merge**.

julesmcrt (approved):

julesmcrt (approved):

PR #558, opened 2026-07-17
feat(builtins): implement systemctl

matt-dz:
## Summary

- Add a restricted `systemctl` builtin with exactly eight commands: `list-units`, `status`, `start`, `stop`, `reload`, `restart`, `enable`, and `disable`.
- Require remediation mode for the entire `systemctl` surface, including listing, status inspection, mutations, and `--help`.
- Enforce exact per-unit action grants through the shared `AllowedSystemServices` policy.
- Add a bounded pure-Go systemd D-Bus backend without executing host `systemctl` or exposing a generic D-Bus client.

## Why

Agents need a controlled way to inspect and remediate explicitly authorized systemd units. The command-wide remediation gate keeps all manager access behind an explicit operator opt-in, while exact unit/action grants constrain which units are visible and which operations are permitted.

## Required configuration

Every invocation must pass all three policy gates:

| Control | Required configuration | Change in this PR |
|---|---|---|
| Command authorization | Include `rshell:systemctl` in `AllowedCommands`, unless all commands are explicitly allowed | Reuses the existing command policy |
| Execution mode | Set `interp.WithMode(interp.ModeRemediation)` | Reuses the existing mode; every `systemctl` invocation now requires it |
| Unit allowlist | Add exact, fully suffixed unit/action grants through `AllowedSystemServices` | Extends the existing policy for restricted `systemctl` |

The existing `read`, `reload`, and `restart` grants are reused. This PR adds `start`, `stop`, `enable`, and `disable`. The existing `clean` grant remains specific to bounded `journalctl` cleanup and does not enable host `systemctl clean`.

The development CLI accepts the equivalent configuration through:

- `--allowed-commands rshell:systemctl`
- `--mode remediation`
- `--allowed-services UNIT:ACTION[+ACTION...]`

## Host access requirements

The process embedding rshell needs access to the target system manager:

| Resource | Standard local path | Required access | Change in this PR |
|---|---|---|---|
| Machine ID | `/etc/machine-id` | Read access for target validation | Existing `MachineIDPath`, reused from `journalctl` |
| Public system D-Bus socket | `/run/dbus/system_bus_socket` | Connect access and noninteractive authorization for the requested systemd methods | New `ManagerBusSocket` field and `--systemd-manager-socket` flag |
| Procfs descriptor links | `/proc/self/fd` | Procfs mounted and accessible for pinned-socket connections | Existing requirement for journal rotation; required for all manager connections |

No explicit `SystemdTargetConfig` is needed for the local host when these standard paths are available. A container or mounted-host integration supplies the container-visible absolute paths:

```go
interp.WithSystemdTarget(interp.SystemdTargetConfig{
	MachineIDPath:    "/host/etc/machine-id",
	ManagerBusSocket: "/host/run/dbus/system_bus_socket",
})
```

Once any explicit target field is configured, omitted fields do not fall back to local paths. `MachineIDPath` remains mandatory, and `ManagerBusSocket` must be supplied for manager-backed `systemctl` operations. Both resources must come from the same host.

The process authenticates to the public system bus with its effective UID and must already be authorized by the host D-Bus, systemd, and any applicable Polkit policy. Rshell performs no privilege elevation or interactive authorization. The private `/run/systemd/private` socket is unsupported.

`JournalDirs` and `JournalControlSocket` are not required solely for `systemctl`. The configured machine ID and manager socket are trusted target paths that bypass `AllowedPaths`. `/proc/self/fd` is a fixed backend path that is also accessed directly; it is not configurable through `ProcPath`.

The configured manager endpoint must be a real Unix socket rather than a final symlink. Missing socket or procfs access, D-Bus authentication or authorization denial, and a machine-ID mismatch all fail closed.

## Behavior and security boundaries

- `list-units` considers only units carrying an exact `read` grant.
- `status` exposes only bounded state fields and omits logs, process command lines, arbitrary properties, unit-file paths, and D-Bus object paths.
- `start`, `stop`, `reload`, and `restart` use fixed synchronous manager jobs and require the matching exact action grant.
- `enable` and `disable` require their matching exact grant, may follow installation metadata, and perform an implicit global manager reload.
- Systemd may still operate on dependencies of an authorized unit as part of its normal transaction semantics.
- `show`, `is-active`, `is-failed`, `is-enabled`, `try-restart`, `reload-or-restart`, `try-reload-or-restart`, `reset-failed`, and `--now` are intentionally unavailable.
- Generic D-Bus methods, arbitrary properties, globs, user/machine/root targets, power verbs, and standalone `daemon-reload` remain unavailable.
- Manager-backed operations require Linux, but do not require cgo, `libsystemd`, or the host `systemctl` executable.

## Validation

- `make fmt`
- `go test ./...`
- `go test -race ./internal/systemd`
- `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash -timeout 120s`
- Linux and Windows cross-compilation checks for the systemd backend and `systemctl` packages
- Independent security audit of the implementation and command-wide remediation gate


matt-dz:
@codex review

builtins/systemctl/systemctl.go
```diff
@@ -0,0 +1,978 @@
...
+	dot := strings.LastIndexByte(unit, '.')
+	return dot > 0 && unit[dot-1] == '@'
+}
+
+func activeStateSuccess(state string) bool {
+	return state == "active" || state == "reloading" || state == "refreshing"
+}
+
+func enabledStateSuccess(state string) bool {
+	return state == "enabled" || state == "enabled-runtime" || state == "static" || state == "alias" || state == "indirect" || state == "generated"
```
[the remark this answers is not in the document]
matt-dz: Not changing this one. The upstream man-page table does list `transient` with exit code 0, but the actual `verb_is_enabled` implementation excludes it from both success sets: `enabled`, `enabled-runtime`, `static`, `alias`, `indirect`, and `generated` are the only successful states. That is true in systemd v252, v255, v258, v260, and current main: https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl-is-enabled.c. Rshell intentionally matches the observable native `systemctl` behavior here; adding `transient` would diverge from it. This appears to be an upstream documentation/source inconsistency.

matt-dz:
@codex review

builtins/systemctl/systemctl.go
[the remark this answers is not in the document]
matt-dz: This duplicates the prior thread, and I revalidated it against the native implementation. systemctl-is-enabled.c treats enabled, enabled-runtime, static, alias, indirect, and generated as successful states; transient is intentionally absent from that switch. rshell follows the behavior of the executable here, while the systemctl manual table currently conflicts with the source. No code change is warranted. Upstream implementation: https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl-is-enabled.c

matt-dz:
@codex review

cmd/rshell/main.go
```diff
@@ -162,13 +164,14 @@ func run(ctx context.Context, args []string, stdin io.Reader, stdout, stderr io.
...
+	cmd.Flags().StringVar(&allowedServices, "allowed-services", "", "comma-separated systemd unit grants in UNIT:ACTION[+ACTION...] form")
 	cmd.Flags().BoolVar(&allowAllCmds, "allow-all-commands", false, "allow execution of all commands (builtins and external)")
 	cmd.Flags().DurationVar(&timeout, "timeout", 0, "maximum execution time for the entire shell run (e.g. 100ms, 5s, 1m)")
 	cmd.Flags().StringVar(&procPath, "proc-path", "", "path to the proc filesystem used by ps (default \"/proc\")")
 	cmd.Flags().StringVar(&journalDirs, "systemd-journal-dirs", "", "comma-separated journal root directories for an explicit systemd target")
 	cmd.Flags().StringVar(&machineIDPath, "systemd-machine-id-path", "", "machine-id file for an explicit systemd target")
 	cmd.Flags().StringVar(&journalSocket, "systemd-journal-socket", "", "journald Varlink socket for an explicit systemd target")
+	cmd.Flags().StringVar(&managerSocket, "systemd-manager-socket", "", "system D-Bus socket for an explicit systemd target")
 	cmd.Flags().StringVar(&mode, "mode", "read-only", "shell execution mode: read-only (default) or remediation (enables file-target output redirections within :rw AllowedPaths roots)")
```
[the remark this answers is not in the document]
matt-dz: Fixed in 376a1047. The --mode help now states that remediation enables file-target output redirections and remediation-only builtins, including the restricted systemctl builtin. TestHelp now asserts that wording. Validation passed: make fmt; go test ./cmd/rshell -run '^TestHelp$' -count=1; go test ./cmd/rshell; gofmt -l .; git diff --check.

internal/systemd/manager_protocol.go
```diff
@@ -0,0 +1,730 @@
...
+
+func validateManagerObjectPath(name string, path dbus.ObjectPath, prefix string) error {
+	if len(path) > maxManagerObjectPath {
+		return fmt.Errorf("systemd manager returned an oversized %s object path", name)
+	}
+	if !path.IsValid() || !strings.HasPrefix(string(path), prefix) || len(path) == len(prefix) {
+		return fmt.Errorf("systemd manager returned an invalid %s object path", name)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
matt-dz: Fixed in a21e4ec0. validateManagerObjectPath now requires exactly one nonempty suffix component after the expected prefix and rejects any additional slash-separated segment. TestManagerBoundaryValidation covers both a valid direct unit leaf and a nested path rejection. Validation passed: make fmt; focused boundary test; go test ./internal/systemd; go test -race ./internal/systemd; go test ./...; and the Docker-backed Bash comparison suite.

AlexandreYang (commented):
## Review Summary

Reviewed the complete `systemctl` builtin change, including exact-unit/action authorization, remediation-mode gating, systemd target configuration, pinned public D-Bus socket access, EXTERNAL authentication and machine-ID verification, bounded message/signal handling, fixed manager methods and properties, partial-progress errors, cross-platform stubs, documentation, and tests.

**Overall assessment: safe to merge.** I found no actionable P0–P3 issues.

| — | — | — | No actionable findings |

### Coverage summary

| Code path | Scenario test | Go test | Status |
|-----------|:---:|:---:|--------|
| remediation gate and default-deny grants | yes | yes | Covered |
| bounded list/status/show/is-* behavior | partial (restricted scenarios) | yes | Covered |
| job/reset/enable/disable authorization and partial progress | — | yes | Covered |
| D-Bus framing, machine identity, inbound calls, and signal overflow | — | yes | Covered |
| unsupported non-Linux backend | CI platform builds/tests | yes | Covered |

Overall coverage: **Adequate**. Scenario coverage is necessarily limited because the harness has no configured systemd manager; backend and happy-path behavior is exercised with focused Go tests. GitHub CI passes tests on Linux, macOS, and Windows, Bash comparison, fuzz suites, CodeQL, gofmt, and analysis. The only failing check is the manual `Static Analysis Label Check`; per repository policy, reviewers must not add `verified/allowed_symbols`.

### Positive observations

- Complete authorization is performed before effects, including compound verbs and `--now`.
- The backend exposes structured capabilities only and fixes destinations, object paths, interfaces, methods, properties, and job mode.
- The manager socket is final-symlink-safe and inode-pinned; the peer machine ID is verified before manager calls.
- Incoming D-Bus messages, authentication input, signals, fields, unit counts, and execution time are explicitly bounded.
- Partial and uncertain mutation outcomes are reported without implying rollback.
- User-facing documentation and builtin help describe the capability and indirect-effect model clearly.

Local test execution was blocked before compilation because the configured internal Go toolchain mirror was not resolvable from this environment; the passing GitHub CI runs provide the test evidence above.

builtins/builtins.go
```diff
@@ -58,14 +58,21 @@ type AllowedPath struct {
...
-	SystemServiceRestart SystemServiceAction = "restart"
+	SystemServiceRead        SystemServiceAction = "read"
+	SystemServiceClean       SystemServiceAction = "clean"
+	SystemServiceStart       SystemServiceAction = "start"
+	SystemServiceStop        SystemServiceAction = "stop"
+	SystemServiceReload      SystemServiceAction = "reload"
+	SystemServiceRestart     SystemServiceAction = "restart"
+	SystemServiceResetFailed SystemServiceAction = "reset-failed"
+	SystemServiceEnable      SystemServiceAction = "enable"
+	SystemServiceDisable     SystemServiceAction = "disable"
```
AlexandreYang: Shall we put those info in a separate system_service.go file? to keep `builtins/builtins.go` cleaner

AlexandreYang: Could be potentially about moving things to builtins/systemd.go 

and renaming builtins/systemd.go to builtins/system_service.go ?

builtins/builtins.go
```diff
@@ -58,14 +58,21 @@ type AllowedPath struct {
...
+// retained for API compatibility; grants may name any exact unit type, such
+// as .service, .timer, or .socket.
 type SystemServiceAction string
 
 const (
-	SystemServiceRead    SystemServiceAction = "read"
-	SystemServiceClean   SystemServiceAction = "clean"
-	SystemServiceReload  SystemServiceAction = "reload"
-	SystemServiceRestart SystemServiceAction = "restart"
+	SystemServiceRead        SystemServiceAction = "read"
```
AlexandreYang: Do we need to support all those actions/verbs right away? or can we keep it more focus/lean at first?
Are those verbs all used in practice? (wondering if some verbs might be deprecated/legacy)

I'm thinking that is we also use this "system service" for other platforms like windows/macOS, it might be easier to have something cross platform is we only focus on important actions first.

julesmcrt: Agree that we should start with a very narrow list of verbs

internal/systemd/manager_linux.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (c *Client) dialManagerBus(ctx context.Context, path string) (net.Conn, error) {
+	socket, err := c.openTargetFileFlags(path, unix.O_PATH|unix.O_NOFOLLOW)
+	if err != nil {
+		return nil, fmt.Errorf("inspect systemd manager bus socket: %w", err)
+	}
+	defer socket.Close()
+	info, err := socket.Stat()
+	if err != nil {
```
AlexandreYang, line 186: <sub><sub>P2 Badge</sub></sub> **Exercise the real pinned-socket transport on Linux**

The security-critical claims in this function are not covered by any test: there is no Linux test that creates a Unix listener, verifies a normal `/proc/self/fd/<O_PATH fd>` connection, rejects a final symlink, and proves that renaming/replacing the configured pathname after pinning cannot redirect the dial. `openManagerBus` authentication/Hello, peer machine-ID verification over a real transport, deadline cancellation, and descriptor cleanup are also only tested below this layer with a fake `managerBus`.

This matters because `/proc/self/fd` reopening is subtle and has had recent security failures when combined with `O_PATH|O_NOFOLLOW` (for example CVE-2026-13201). The explicit `fstat` socket check appears to reject that CVE’s final-symlink case here, but a Linux integration test should lock in that property and the path-replacement guarantee rather than leaving the central trust-boundary assertion unverified.

Please add Linux-only transport tests with a temporary Unix socket and a synchronization hook/helper between pin and dial, covering: real socket success, regular-file rejection, final-symlink rejection, pathname replacement after pinning, context deadline, and fd cleanup. A small raw/fake D-Bus server test should additionally cover Auth + Hello + matching/mismatching `Peer.GetMachineId` end to end.

AlexandreYang (commented):
## Deep security review

**Assessment: needs one test-hardening fix before I would call this high-confidence safe.** I found no demonstrated sandbox escape, authorization bypass, command injection, arbitrary D-Bus surface, or unbounded resource path. The documented trust model is internally consistent, but the most security-sensitive Linux transport claims are currently established only by code inspection.

| 1 | <sub><sub>P2 Badge</sub></sub> | `internal/systemd/manager_linux.go:201` | The actual pinned-socket transport and path-replacement guarantees have no Linux test |

### Threat-model conclusions

- Shell input cannot select paths, destinations, object paths, interfaces, methods, property names, or job modes.
- Every explicit unit/action pair is validated and authorized before any effect; compound verbs and `--now` preauthorize the full operation.
- Alias canonicalization is output-only and does not expand grants.
- Dependencies, install metadata, and manager reloads can affect other units, but this is explicitly documented as trusted systemd behavior rather than accidental policy expansion.
- The D-Bus client bounds authentication input, individual messages, fields, signals, unit counts, and total operation time.
- A final symlink at `ManagerBusSocket` is rejected: `O_PATH|O_NOFOLLOW` pins the link itself and the subsequent `fstat` fails the socket-type check. This avoids the direct leaf-symlink pattern described by CVE-2026-13201.
- Parent path components and the target machine-ID file remain trusted operator configuration. If an attacker can replace those components, they can potentially present a fake bus and repeat the public machine ID; this is outside the documented shell-attacker boundary and should remain explicit in deployment guidance.
- The configured bus daemon/systemd peer is a trusted host component. A malicious bus could still exercise dependency decoder behavior (for example malformed bus-control signals), but an rshell script cannot produce those frames on a correctly configured system bus.

### Operational safety caveat

An exact grant is not a containment boundary around one unit. Starting a trusted unit may activate dependencies; `enable`/`disable` may follow `Alias=`, `Also=`, or `DefaultInstance=` and then globally reload the manager. A grant to a target or lifecycle-sensitive unit can therefore reboot, shut down, or substantially reconfigure the host. The PR documents this accurately, but operators must treat granted unit payloads and dependency graphs as trusted.

### Coverage

| Code path | Test | Status |
|-----------|:---:|--------|
| remediation gate/default deny | scenario + Go tests | Covered |
| exact validation and authorization ordering | `systemctl_test.go`, `system_services_test.go` | Covered |
| fixed manager calls/reply validation/partial progress | fake-bus protocol tests | Covered |
| D-Bus size/auth/signal bounds | focused unit tests | Covered |
| real Linux `O_PATH` socket pin and `/proc/self/fd` dial | — | **Missing** |
| final-symlink and path-replacement attack cases | — | **Missing** |
| real auth/Hello/machine-ID/timeout cleanup | — | **Missing** |
| non-Linux fail-closed behavior | platform CI/build | Partially covered |

GitHub CI passes the platform tests, Bash comparison, fuzz suites, CodeQL, formatting, and analysis. Local full tests were unavailable because the configured internal Go toolchain mirror could not resolve. OSV reports no known advisory for `github.com/godbus/dbus/v5` v5.2.2 as of this review.

AlexandreYang (commented):
## Code coverage review

**Overall coverage: gaps found.** The builtin/parser and authorization layers are well covered, but important backend and platform behavior is either mock-only or untested. This review is a `COMMENT`, not an approval.

| 1 | <sub><sub>P2 Badge</sub></sub> | `internal/systemd/manager_linux.go:201` | Real pinned-socket transport, replacement resistance, D-Bus handshake, deadlines, and cleanup have no Linux test (reported in the prior deep-security review) |
| 2 | <sub><sub>P2 Badge</sub></sub> | `internal/systemd/manager_protocol.go:348` | `DisableUnitFiles` and persistent-mutation failure semantics are not covered |
| 3 | <sub><sub>P2 Badge</sub></sub> | `internal/systemd/manager_protocol.go:425` | Most job-completion fail-closed branches are not covered |
| 4 | <sub><sub>P3 Badge</sub></sub> | `internal/systemd/manager_unsupported.go:17` | Non-Linux stubs are compiled but not behaviorally tested |

### Coverage inventory

| Code path | Test | Status |
|-----------|:---:|--------|
| command-wide remediation gate | scenario + `TestSystemctlRequiresRemediationModeBeforeAnyCapability` | Covered |
| exact operand validation and limits | `TestExplicitOperandsAreValidatedBeforeAuthorization`, `TestArgumentTokenCountsAreBoundedBeforeBackendWork` | Covered |
| restricted list construction/filtering/backend over-return | list-focused builtin tests | Covered |
| status/show/is-* formatting and principal exit codes | builtin tests | Covered |
| fixed job verb/action mapping and compound preauthorization | `TestJobVerbsUseFixedBackendActionsAndAuthorizations`, compound tests | Covered |
| enable/disable `--now` ordering and template rejection | builtin fake-controller tests | Covered |
| backend inspection/property selection and reply validation | fake-bus protocol tests | Mostly covered |
| successful and partial runtime jobs | protocol tests | Covered |
| signal overflow and ambiguous job delivery | protocol tests | Covered |
| context cancellation, closed/nil/malformed signals, unsuccessful job results, unrelated-signal cap | — | **Missing** |
| `EnableUnitFiles` success + reload | `TestEnableSystemServicesReloadsManagerAndBoundsChanges` | Covered |
| `DisableUnitFiles` request/reply/reload | — | **Missing** |
| enable/disable D-Bus errors, malformed replies, invalid changes, reload errors | — | **Missing** |
| reset-failed successful backend call and malformed reply | — | **Missing** |
| real Linux socket pin/auth/Hello/machine-ID/deadline cleanup | — | **Missing** |
| D-Bus frame/auth byte limits and signal queue | focused tests | Covered |
| non-Linux methods return `ErrSystemdUnsupported` | — | **Missing** |
| CLI target configuration and documentation/help | CLI/target/scenario tests | Covered |

### Positive observations

- Tests explicitly prove authorization completes before effects.
- Security-sensitive output sanitization and backend over-return checks are covered.
- Fixed D-Bus methods/properties and size/count bounds receive focused unit coverage.
- Scenario tests correctly cover the restricted behavior available without a real systemd target.

The missing cases should be covered primarily with Go tests: scenario tests cannot conveniently provide a real or fake manager bus. GitHub CI currently passes platform suites, Bash comparison, fuzzing, CodeQL, analysis, and formatting; that does not execute the missing transport or negative backend branches.

internal/systemd/manager_protocol.go
```diff
@@ -0,0 +1,750 @@
...
+		return unitFilePartialProgress(err)
+	}
+	return nil
+}
+
+func disableSystemServicesWithBus(ctx context.Context, bus managerBus, units []string) error {
+	if err := validateManagerUnits(units, false); err != nil {
+		return err
+	}
+	body, err := bus.call(ctx, systemdBusDestination, systemdManagerPath, systemdManagerIface+".DisableUnitFiles", units, false)
```
AlexandreYang, line 290: <sub><sub>P2 Badge</sub></sub> **Cover disable and persistent-mutation failure semantics**

The protocol suite tests only the successful `EnableUnitFiles` path. There is no test that calls `disableSystemServicesWithBus`, so its D-Bus method name/signature (`DisableUnitFiles`, units, `false`), reply decoding, change validation, and required manager reload could regress without detection. The mutation error branches shared by enable/disable are also uncovered: D-Bus rejection/transport ambiguity, malformed replies, unsupported or duplicate change records, and reload error/malformed reply with the required “changes completed / not rolled back” warning.

Please add table-driven fake-bus tests for enable and disable covering success plus each failure phase. Assert the exact call arguments and ordering, and assert that failures after the unit-file call retain the partial-progress warning. Also cover a successful `ResetFailedUnit` backend call and its malformed-empty-reply branch, which are currently only exercised at the builtin/fake-controller layer.

internal/systemd/manager_protocol.go
```diff
@@ -0,0 +1,750 @@
...
+		return "", false
+	}
+}
+
+func managerTryMethodIgnoresMasked(method string, err error) bool {
+	return managerDBusErrorName(err) == "org.freedesktop.systemd1.UnitMasked" &&
+		(method == "TryRestartUnit" || method == "ReloadOrTryRestartUnit")
+}
+
+func waitSystemdJob(ctx context.Context, signals <-chan *dbus.Signal, overflow <-chan struct{}, expectedPath dbus.ObjectPath, selector string) error {
```
AlexandreYang, line 352: <sub><sub>P2 Badge</sub></sub> **Exercise every job-wait terminal path**

Current tests cover a matching `done`/`skipped` signal and queue overflow, but not most of this fail-closed state machine: canceled context, closed signal channel, nil signal, malformed body, mismatched job ID/path, invalid unit/result strings, explicit failed/canceled/timeout job results, ignored unrelated signals, or the `maxManagerSignalsRead` exhaustion guard. These branches decide whether a host mutation is reported as successful, failed, or outcome-unknown, so regressions are operationally significant.

Add a table-driven `waitSystemdJob` test that constructs each terminal input and asserts its exact classification. Include at least one unrelated signal followed by the matching signal, and `maxManagerSignalsRead` unrelated signals to verify the hard cap.

internal/systemd/manager_unsupported.go
```diff
@@ -0,0 +1,47 @@
...
+package systemd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/DataDog/rshell/builtins"
+)
+
+func (*Client) ListSystemServices(context.Context, builtins.SystemServiceListRequest) ([]builtins.SystemServiceState, error) {
```
AlexandreYang, line 17: <sub><sub>P3 Badge</sub></sub> **Assert non-Linux methods fail closed**

macOS and Windows CI compile and run the package, but no test calls these stubs. Add a `//go:build !linux` test that invokes every reader/controller method and checks `errors.Is(err, builtins.ErrSystemdUnsupported)`, with no returned state. This locks in the documented fail-closed platform contract.

builtins/systemctl/systemctl.go
```diff
@@ -0,0 +1,978 @@
...
+		case "stop":
+			return runJobWithFlags(ctx, callCtx, fs, verb, operands, builtins.SystemServiceJobStop, builtins.SystemServiceStop)
+		case "reload":
+			return runJobWithFlags(ctx, callCtx, fs, verb, operands, builtins.SystemServiceJobReload, builtins.SystemServiceReload)
+		case "restart":
+			return runJobWithFlags(ctx, callCtx, fs, verb, operands, builtins.SystemServiceJobRestart, builtins.SystemServiceRestart)
+		case "try-restart":
+			return runJobWithFlags(ctx, callCtx, fs, verb, operands, builtins.SystemServiceJobTryRestart, builtins.SystemServiceRestart)
+		case "reload-or-restart":
+			return runJobWithFlags(ctx, callCtx, fs, verb, operands, builtins.SystemServiceJobReloadOrRestart, builtins.SystemServiceReload, builtins.SystemServiceRestart)
```
julesmcrt: <sub><sub>P3 Badge</sub></sub> **`reload-or-restart` requires both `reload` and `restart` grants**

`reload-or-restart`/`try-reload-or-restart` pass both `SystemServiceReload` and `SystemServiceRestart` to `authorize()`, which requires every unit×action pair (line 549-563). So an operator who only wants to grant "restart" can't use `reload-or-restart` without also granting `reload`, even though systemd only ever performs one of the two per invocation. This fails closed (not a bypass) — flagging as a least-privilege/usability gap rather than a bug. If intentional (e.g. because the caller can't know in advance which job systemd will pick), a short comment here would help future readers.

internal/systemd/manager_dbus_limit.go
```diff
@@ -0,0 +1,129 @@
...
+	header     [16]byte
+	headerOff  int
+	remaining  uint64
+}
+
+func (c *boundedDBusConn) Write(data []byte) (int, error) {
+	if bytes.Equal(data, dbusAuthBegin) {
+		// Auth writes BEGIN immediately before starting the binary message reader.
+		// Publish the mode first so a fast bus reply cannot race the transition.
+		c.binaryMode.Store(true)
```
julesmcrt, line 44: <sub><sub>P3 Badge</sub></sub> **`binaryMode` switch depends on godbus writing `BEGIN\r\n` as an isolated `Write` call**

`c.binaryMode.Store(true)` only fires on an *exact* match against `dbusAuthBegin` in a single `Write`. This is correct today (verified godbus issues `BEGIN\r\n` as its own write), and it's not attacker-influenced since rshell drives both sides of the auth exchange — but it's an implicit coupling to godbus's internal write pattern. If a future godbus version ever batches `BEGIN\r\n` with adjacent bytes in one `Write`, the bound would silently stay in unbounded auth mode. Worth a comment noting the dependency, plus an adversarial test case (`BEGIN\r\n` glued to other bytes) alongside the existing happy-path test in `manager_dbus_limit_test.go:47-64`.

builtins/systemctl/systemctl.go
```diff
@@ -0,0 +1,978 @@
...
+				return result
+			}
+			return runResetFailed(ctx, callCtx, operands)
+		case "enable", "disable":
+			if result, ok := rejectFlags(callCtx, fs, verb, "now"); !ok {
+				return result
+			}
+			return options.runEnableDisable(ctx, callCtx, verb, operands)
+		default:
+			callCtx.Errf("systemctl: unsupported command %q\n", safeText(verb))
```
julesmcrt, line 106: <sub><sub>P3 Badge</sub></sub> **Untested: rejection of entirely-unimplemented verbs**

Dangerous *flags* (`--user`, `--machine`, `--global`, `--runtime`, `--force`, `--job-mode`, `--root`) have an explicit rejection test, but no test (Go or scenario) exercises this `default:` branch for verbs that don't exist at all in the switch — `poweroff`, `reboot`, `daemon-reload`, `mask`, `unmask`, `kill`, `edit`, `cat`, `list-unit-files`, `isolate`. They all correctly fall through to `unsupported command %q`, but the branch itself isn't locked in by a test. Suggest a small table-driven Go test asserting exit code 1 + the exact stderr message for a few of these.

builtins/systemctl/systemctl.go
```diff
@@ -0,0 +1,978 @@
...
+		seen[unit] = struct{}{}
+		units = append(units, unit)
+		if len(units) > builtins.MaxSystemServiceOperands {
+			return units
+		}
+	}
+	return units
+}
+
+func validateUnitName(unit string) error {
```
julesmcrt, line 394: <sub><sub>P3 Badge</sub></sub> **No explicit path-traversal-shaped unit name test**

`glob_rejected.yaml` covers `api*.service`, but there's no test asserting rejection of a path-traversal-shaped operand like `../../etc/passwd.service` or `foo/bar.service`. Almost certainly already rejected here (no `/` allowed), but given this function is the one thing standing between shell input and D-Bus object-path construction, an explicit test would make that guarantee visible rather than incidental.

julesmcrt (commented):
## Review Summary

Reviewed the restricted `systemctl` builtin, the new pure-Go systemd D-Bus manager backend (`internal/systemd/manager_*.go`), and the `AllowedSystemServices`/remediation-mode authorization gates, focused on sandbox escape, authorization bypass, and DoS.

**Overall assessment: safe to merge.** No P0/P1 issues found. I cross-checked against the existing threads from @chatgpt-codex-connector, @Copilot, and @AlexandreYang first to avoid duplicating already-open items (pinned-socket transport test gap, `disable`/persistent-mutation test gap, job-wait terminal-state test gap, non-Linux stub test gap, and the `is-enabled`/`transient` divergence are all already tracked in open threads above — not re-filed here). The handful of net-new items below are all P2/P3 hardening or test-coverage notes, nothing blocking.

### Positive observations
- Socket pinning in `dialManagerBus` (`O_PATH|O_NOFOLLOW` + `fstat` + dial via `/proc/self/fd/<fd>`) correctly eliminates the TOCTOU between path validation and connect.
- Machine-ID verification has no bypassable early-return; mismatches/malformed IDs fail closed.
- Context cancellation is enforced end-to-end (godbus's own cancellation goroutine plus a matching socket deadline), so a hung D-Bus peer can't block past the 30s exec timeout.
- `list-units` strictly enumerates from `ReadableSystemServices()` — no way to probe for ungranted unit names via `status`/`show`/`is-*`, all of which authorize before touching the backend.
- `show` is locked to a fixed property allowlist; no generic D-Bus methods, globs, `--user`/`--machine`/`--global`/`--root`, power verbs, or standalone `daemon-reload` are reachable.
- Compound operations (`enable --now`, `disable --now`, `reload-or-restart`) authorize every required unit/action pair before any effect fires.

### Additional note (no good diff anchor for an inline comment)
AGENTS.md's "Testing" section documents scenario-test limitations for `ip route`/`free` (Linux-only commands the scenario framework can't express), but doesn't yet note the analogous gap for `systemctl`: the scenario runner has no field for configuring `AllowedSystemServices` grants, so every authorized-success path (`list-units` with grants, `status`, `show`, `is-*`, job verbs, `enable`/`disable`) can only be scenario-tested for the denied case — authorized-success coverage necessarily lives in `builtins/systemctl/systemctl_test.go`'s fakes instead. Worth a one-line addition alongside the existing `ip route`/`free` exceptions for consistency, since that section isn't touched by this diff.

builtins/builtins.go
matt-dz: i can remove `reset-failed` and `clean`. i think the rest of the verbs are basic and useful systemctl commands (`start`, `stop`, `reload`, `restart`, `enable`, `disable`)

internal/systemd/manager_linux.go
matt-dz: @datadog fix this

builtins/systemctl/systemctl.go
matt-dz: seems *very* nitpicky -- not planning on doing this.

builtins/systemctl/systemctl.go
matt-dz: addressed in new tests

internal/systemd/manager_dbus_limit.go
matt-dz: addressed

internal/systemd/manager_unsupported.go
matt-dz: addressed with new tests

internal/systemd/manager_protocol.go
matt-dz: addressed with new tests

internal/systemd/manager_protocol.go
matt-dz: address with new tests

builtins/builtins.go
matt-dz: I’ve narrowed it to eight commands:

- Inspection: `list-units`, `status`
- Runtime lifecycle: `start`, `stop`, `reload`, `restart`
- Boot-time configuration: `enable`, `disable`

I removed show, the `is-*` predicates, `try-restart`, the compound `reload`/`restart` variants, `reset-failed`, and `--now`. I also removed their backend methods, action grants, and API types rather than merely hiding the CLI verbs.

The retained lifecycle operations are common and non-legacy. The current implementation remains explicitly Linux/systemd-only and fails closed elsewhere, but this smaller surface gives us a much better starting point if we later introduce Windows or macOS service-manager backends. We would still need to define platform-specific semantics—especially for reload and enablement—instead of assuming they map identically.

matt-dz:
@codex review

astuyve (approved):

PR #562, opened 2026-07-22
feat(builtins): implement uptime builtin

san-jos:
## Summary

- Adds the \`uptime\` builtin command with cross-platform support (Linux, macOS, Windows)
- Default output: current time, uptime duration, and load averages (1/5/15 min); user count intentionally omitted (reading utmp would expose session data)
- Flags: \`-p\`/\`--pretty\` (human-readable duration), \`-s\`/\`--since\` (boot time as \`YYYY-MM-DD HH:MM:SS\`), \`-h\`/\`--help\`

## Data sources

All bypass \`AllowedPaths\` (fixed kernel paths — same precedent as \`ss\`, \`ip route\`, \`df\`):

| Platform | Uptime | Load avg |
|----------|--------|----------|
| Linux | \`/proc/uptime\` | \`/proc/loadavg\` |
| macOS | \`sysctl kern.boottime\` | \`sysctl vm.loadavg\` |
| Windows | \`GetTickCount64\` via kernel32.dll | not available |

Windows 32-bit (386): \`GetTickCount64\` returns a 64-bit ULONGLONG split across lo (EAX) and hi (EDX) — combined in \`sysinfo_windows_386.go\`. On 64-bit targets hi is XMM0 (unspecified) and is ignored in \`sysinfo_windows_64bit.go\`.

## Test plan

- [x] Unit tests: \`builtins/uptime/uptime_test.go\` — \`formatDuration\`, \`formatPretty\`, \`formatDefault\` with deterministic fixed inputs (no OS calls)
- [x] Integration tests: \`builtins/tests/uptime/uptime_test.go\` — 6 handler-level cases with injected fake provider (default output, \`-p\`, \`-s\`, \`-p\` beats \`-s\`, generic error, \`ErrNotSupported\`)
- [x] Scenario tests: flag rejection, extra operands, \`--help\`, \`--\` end-of-flags separator (error cases only; happy-path scenarios omitted as \`sysinfo.Get()\` may return EPERM in restricted CI environments)
- [x] Cross-compile verified: \`GOOS=windows GOARCH={amd64,386,arm64}\` all build clean
- [x] All analysis/symbol allowlist tests pass


builtins/internal/sysinfo/sysinfo_other.go
matt-dz, line 1: nit: file may be redundant as rshell doesn't aim to target systems outside of linux, darwin, or windows

san-jos:
@codex review

matt-dz (approved):

builtins/internal/sysinfo/sysinfo_windows_64bit.go
```diff
@@ -0,0 +1,15 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-present Datadog, Inc.
+
+//go:build windows && !386
```
julesmcrt, line 6: <sub><sub>P2 Badge</sub></sub> **Build tag `windows && !386` also matches `windows/arm`, which likely needs the lo/hi combine logic instead**

This file's comment justifies discarding `hi` because it's "the XMM0 floating-point register (unspecified for integer-returning functions)" — that's an x86-64/ARM64 calling-convention fact, not one that holds for 32-bit ARM. `windows && !386` matches `windows/amd64`, `windows/arm64`, **and `windows/arm`**. On 32-bit ARM, a `uint64` return from a stdcall-style DLL call is conventionally split across two registers the same way it is on `windows/386`, so building for `windows/arm` would silently truncate `GetTickCount64` to 32 bits and reintroduce the ~49.7-day rollover this PR just fixed for x86.

The PR's own cross-compile check list (`amd64`, `386`, `arm64`) doesn't include `arm`, so this wasn't caught.

Suggest changing the tag to `windows && !386 && !arm` and adding an `arm`-specific `ticksToMs` (same combine logic as 386), or explicitly documenting `windows/arm` as unsupported if it's out of scope for this project.

builtins/internal/sysinfo/sysinfo_windows_386.go
```diff
@@ -0,0 +1,13 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-present Datadog, Inc.
+
+package sysinfo
+
+// ticksToMs reconstructs the full ULONGLONG from a Proc.Call result on
+// 32-bit Windows. The Win32 calling convention returns 64-bit integers with
+// the low 32 bits in EAX (lo) and the high 32 bits in EDX (hi).
+func ticksToMs(lo, hi uintptr) uint64 {
```
julesmcrt, line 11: <sub><sub>P2 Badge</sub></sub> **No unit tests for the `sysinfo` package**

None of `sysinfo_linux.go`, `sysinfo_darwin.go`, or `sysinfo_windows*.go` have accompanying `_test.go` files. A few things stand out:

- `ticksToMs` (this file and `sysinfo_windows_64bit.go`) is a pure function with zero OS dependency — trivially unit-testable — and this exact logic was the target of a dedicated bug-fix commit earlier in this PR's history, which is usually a strong signal it needs a regression test.
- The Linux `/proc/uptime`/`/proc/loadavg` field parsers (`readUptimeSeconds`, `readLoadAvg`) have no test for malformed input (empty file, non-numeric fields, missing columns), even though `docs/RULES.md` requires builtins to "fail safely on malformed... input." The error-handling code path is correct but currently unexercised.
- The Darwin sysctl byte-offset decoders (`readBootTime`, `readLoadAvg`) have no fixture-based test, unlike `diskstats_darwin_test.go` for the comparable `df` internal package.

Sibling internal packages this PR cites as precedent (`meminfo`, `diskstats`) both have dedicated test files with fixed fixtures; `sysinfo` breaking that pattern is a gap worth closing, e.g. a table test for `ticksToMs` on both build variants plus fixture-based tests for the Linux/Darwin parsers.

julesmcrt (commented):
Reviewed the uptime builtin and sysinfo package. No sandbox-escape or security issues found — the AllowedPaths bypass is properly documented following the ss/ip route/df/free precedent, build is clean, and existing tests pass. Two additional (non-blocking) findings below, on top of what's already been raised by matt-dz and codex.

san-jos:
@codex review

san-jos:
@codex review

builtins/internal/sysinfo/sysinfo_windows_64bit.go
san-jos: Good catch on the calling convention — verified it against Go's supported platform list: `go tool dist list | grep windows` shows only `windows/386`, `windows/amd64`, and `windows/arm64`. `windows/arm` (32-bit ARM) is not a supported Go target, so `!386` already covers exactly the right set. Updated the comment in `sysinfo_windows_64bit.go` to make this explicit.

builtins/uptime/uptime.go
```diff
@@ -0,0 +1,181 @@
...
+		MakeFlags: func(fs *builtins.FlagSet) builtins.HandlerFunc {
+			return makeFlags(fs, getInfo)
+		},
+	}
+}
+
+func makeFlags(fs *builtins.FlagSet, getInfo func() (sysinfo.Info, error)) builtins.HandlerFunc {
+	help := fs.BoolP("help", "h", false, "display this help and exit")
+	pretty := fs.BoolP("pretty", "p", false, "show uptime in pretty format")
+	since := fs.BoolP("since", "s", false, "system up since, in yyyy-mm-dd HH:MM:SS format")
```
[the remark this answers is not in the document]
san-jos: This is a valid observation but it applies equally to `cat`, `ping`, `truncate`, `uname`, `logrotate`, `sed`, and other builtins that all use `BoolP` the same way — `df` is currently the only exception that uses `registerNoArgBool`. Fixing it in uptime alone would be inconsistent. Tracking this as a separate codebase-wide cleanup rather than addressing it in this PR.

builtins/internal/sysinfo/sysinfo_windows_386.go
san-jos: Added `sysinfo_linux_test.go` with table tests for `readUptimeSeconds` and `readLoadAvg` covering malformed input (empty file, non-numeric fields, missing columns). `uptimePath`/`loadavgPath` converted from `const` to `var` to allow fixture injection. The Darwin decoder tests would need fixture-based byte slices to test the sysctl parsing — happy to add those in a follow-up if you'd like.

san-jos:
@codex review

builtins/uptime/uptime.go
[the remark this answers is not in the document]
san-jos: Same as the earlier comment — this is a systemic pattern across the codebase (cat, ping, truncate, uname, sed all use BoolP the same way). See reply on the earlier thread. Tracking separately.

tests/scenarios/cmd/uptime/errors/extra_operand.yaml
```diff
@@ -0,0 +1,9 @@
+description: uptime rejects extra positional operands
+skip_assert_against_bash: true
+input:
+  script: |+
+    uptime foo
+expect:
+  stdout: ""
+  stderr: "uptime: extra operand 'foo'\nTry 'uptime --help' for more information.\n"
```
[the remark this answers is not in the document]
san-jos: Fixed in 7892b771 — converted the quoted escape-sequence string to a |+ block scalar.

PR #563, opened 2026-07-22
feat(builtins): add lsof builtin

louis-cqrl:
### What does this PR do?

Adds a pure-Go, Linux-only `lsof` builtin that lists open files per process by reading `/proc/<pid>/fd`. Supports `-p` (PID filter), `-c` (command-name prefix filter), `-u` (UID filter), `-a` (AND the selectors instead of the default OR), and `-h`/`--help`.

### Motivation

`lsof` is the next builtin on the implementation plan, chosen specifically for deleted-open-file diagnostics: surfacing files a process still holds open after they were unlinked (the classic "disk full but `du` shows nothing" root cause).

Like `ss`/`ip route`/`df`/`free`, it bypasses `AllowedPaths` for trusted `/proc/<pid>/fd` metadata reads (COMMAND/PID/USER/FD/TYPE/DEVICE/SIZE/NODE). It deliberately diverges from that precedent for the NAME column: a `/proc/<pid>/fd` symlink target can resolve to any path on the host, so it's gated against `AllowedPaths` and redacted to `(restricted)` (or `(restricted) (deleted)`) outside every configured root. See the new `AGENTS.md` entry for the full rationale, including the memfd-gating edge case.

`lsof` never reads `/proc/<pid>/cmdline` or `/proc/<pid>/environ`, matching `ps`'s argv/environment privacy rule.

### Testing

- `builtins/internal/procfd/procfd_linux_test.go` — unit tests over real `/proc`, including a real-Linux repro of the kernel's `" (deleted)"` disambiguation and a literal-filename-collision regression test.
- `builtins/lsof/lsof_test.go` — selector, path-gating, and UID-canonicalization unit tests.
- `builtins/tests/lsof/` — happy-path, gating, and adversarial pentest tests (`lsof_pentest_linux_test.go`): path-traversal, memfd-naming collision, root-slash, and OR/AND selector semantics.
- `tests/scenarios/cmd/lsof/` — help/error-path scenario tests (happy-path scenarios can't be expressed; same platform-skip limitation as `free`/`ip route`, documented in `AGENTS.md`).
- Fuzz tests wired into `.github/workflows/fuzz.yml`.
- Verified on real Linux (via a container) in addition to macOS: full build, `go vet`, `gofmt`, and the full test suite all pass.

### Checklist

- [x] Tests added/updated
- [x] Documentation updated (`README.md`, `SHELL_FEATURES.md`, `AGENTS.md`)


louis-cqrl:
@codex review

louis-cqrl:
@codex review

builtins/lsof/lsof.go
```diff
@@ -0,0 +1,690 @@
...
+		command: sanitizeField(of.Command),
+		pid:     strconv.Itoa(of.PID),
+		user:    of.UID,
+		fd:      of.FD,
+		typ:     of.Type,
+		device:  of.Device,
+		size:    of.Size,
+		node:    of.Node,
+		name:    sanitizeField(redactName(of, roots, hostPrefix)),
+	}
```
julesmcrt, line 488: `toRow` only routes `Name` through `redactName`; `Device`, `Size`, and `Node` are copied verbatim regardless of whether the path is restricted. For a REG file, `Size` is the exact byte count from `unix.Stat` (see `procfd_linux.go` `readLinkEntry`), so an agent that can't see the *path* of a file outside AllowedPaths can still learn its exact size, device, and inode number — e.g. for `/etc/shadow` or any secret held open by another process.

This is a broader exposure than the `ss`/`df`/`free` precedent this PR cites: those only expose aggregate host-wide counters, never per-file attributes tied to a specific out-of-sandbox path. Given the stated goal of NAME redaction is to avoid "disclosing where on the host filesystem the file lives" while still supporting the deleted-file diagnostic, it seems like the diagnostic use case is actually served fine by blanking SIZE/NODE/DEVICE alongside NAME on restricted rows (COMMAND/PID/USER/FD/TYPE would still show that *something* is open and deleted, without leaking size/inode of an out-of-sandbox file).

If this metadata exposure is intentional, could you call it out explicitly in the AGENTS.md rationale for `lsof`? Right now that doc frames NAME as *the* privacy boundary, which reads as stronger than what the code actually does.

builtins/internal/procfd/procfd_linux.go
```diff
@@ -0,0 +1,345 @@
...
+func readLinkEntry(dir, linkName string, pid int, comm, uid, displayFD string) (OpenFile, bool) {
+	linkPath := filepath.Join(dir, linkName)
+
+	target, err := os.Readlink(linkPath)
+	if err != nil {
+		return OpenFile{}, false
+	}
+
+	var st unix.Stat_t
+	if err := unix.Stat(linkPath, &st); err != nil {
```
julesmcrt, line 220: Minor: `Readlink(linkPath)` (line 207) and `unix.Stat(linkPath, &st)` (here) are two separate syscalls against the same `/proc/<pid>/fd/N` magic symlink. If the target process closes that fd and a new one gets the same number before the `Stat` call runs, `target`/`deleted` come from the old fd but `Type`/`Device`/`Size`/`Node` come from whatever now occupies fd N — a torn read across two different underlying files.

Severity is low (read-only, metadata-only, and bounded by the SIZE/NODE exposure noted separately), but worth flagging since the package doc/PR description describes this as mirroring the journalctl exception's approach, and that exception does re-verify file identity across open (per docs/RULES.md's trusted systemd target section) — this doesn't do the equivalent here. Not necessarily worth fixing (there's no cheap way to atomically read both via `/proc`), but might be worth a short comment acknowledging the race, similar to the existing comment above about the fd-closed-between-scan-and-stat race.

julesmcrt (commented):
Reviewed the implementation in full (procfd package + lsof.go) plus the security-sensitive gating logic specifically. This is a careful, well-documented PR — no critical/high-severity issues, no violations of docs/RULES.md's file-access rules, no cmdline/environ reads, and the AllowedPaths gate on NAME correctly handles `..` traversal, prefix-sibling roots, symlinked roots, and the memfd-collision edge case (all covered by tests). Two things worth a look before merge, left as inline comments:

1. **SIZE/DEVICE/NODE are shown even when NAME is `(restricted)`.** For a regular file this leaks its exact byte size (and device/inode) outside every configured AllowedPaths root — a strictly broader metadata exposure than the `ss`/`df`/`free` precedent, which only exposes aggregate host counters, never per-file attributes for paths outside the sandbox. May be intentional, but worth confirming and calling out explicitly in the AGENTS.md rationale if so.
2. **TOCTOU between `os.Readlink` and `unix.Stat` in `readLinkEntry`.** If the process closes/reuses the fd between the two syscalls, NAME and TYPE/SIZE/NODE can come from two different underlying files. Low severity (read-only, metadata-only, bounded by finding 1's already-accepted exposure), but the PR framing ("mirrors journalctl's approach") implies identity re-verification that isn't actually present here.

Neither blocks approval in my view, but I'd like the author's take on (1) especially, since it changes the actual privacy boundary the redaction is meant to enforce.

matt-dz (approved):

builtins/internal/procfd/procfd_linux.go
```diff
@@ -0,0 +1,346 @@
...
+		// legitimately be named ending in that literal string. Disambiguate
+		// by stat'ing the full, untrimmed target: if a real file exists at
+		// that exact literal path and resolves to the same inode as the
+		// live fd (st, already captured above), the suffix is part of the
+		// actual filename rather than a deletion marker, so it is kept
+		// verbatim and Deleted stays false. Otherwise (no such file, or a
+		// different inode) the suffix is a genuine kernel deletion marker,
+		// so it is stripped and Deleted is set.
+		var literalSt unix.Stat_t
+		if statErr := unix.Stat(target, &literalSt); statErr != nil || literalSt.Dev != st.Dev || literalSt.Ino != st.Ino {
```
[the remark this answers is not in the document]
louis-cqrl: Fixed in f03ad969: removed the extra `stat` call on the readlink target entirely. The ` (deleted)` suffix is now trusted directly from the readlink string without touching the process-controlled path, so this only reads the fixed `/proc/<pid>/fd/<n>` symlink. A comment now documents why the suffix is trusted as-is (a real file literally named ending in ` (deleted)` is indistinguishable and out of scope, matching real `lsof`/`ls`/`readlink` on Linux).

builtins/lsof/lsof.go
```diff
@@ -0,0 +1,690 @@
...
+	flag.NoOptDefVal = noArgSentinel
+	return target
+}
+
+func registerFlags(fs *builtins.FlagSet) builtins.HandlerFunc {
+	pidList := fs.StringP("pid", "p", "", "select by PID list (comma or space separated)")
+	cmdPrefix := fs.StringP("command", "c", "", "select by command-name prefix (literal, no regex)")
+	uidList := fs.StringP("user", "u", "", "select by numeric UID list (comma or space separated)")
+	and := registerNoArgBool(fs, "and", "a", "AND the selection criteria instead of OR")
+	help := registerNoArgBool(fs, "help", "", "print usage and exit")
```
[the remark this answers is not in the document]
louis-cqrl: Fixed in f03ad969: registered `-h` as the shorthand for `--help`, and updated the usage strings/flag docs to `[-h] [--help]`. Added a scenario test (`help_shorthand.yaml`) covering `lsof -h`.

builtins/lsof/lsof.go
louis-cqrl: Good catch, fixed in 9a5137c2. `toRow` now blanks `device`/`size`/`node` on restricted rows using the same boundary check as NAME (extracted into a shared `pathRestricted` helper), so a restricted row only shows COMMAND/PID/USER/FD/TYPE, matching the deleted-file diagnostic use case without exposing per-file byte count/device/inode. Updated the AGENTS.md rationale to describe this explicitly, and added unit tests (`TestToRowBlanksDeviceSizeNodeOnRestrictedRow`, `TestToRowKeepsDeviceSizeNodeWhenAllowed`, `TestToRowKeepsDeviceSizeNodeForNonPathTargets`) plus an end-to-end test (`TestLsofGatingRedactsSizeDeviceNodeOutsideAllowedPaths`) proving a distinctive real file size never appears in stdout for a restricted row.

builtins/internal/procfd/procfd_linux.go
louis-cqrl: Agreed there's no atomic way to do this via /proc. Added a comment above the `Stat` call in 9a5137c2 acknowledging the race explicitly (Readlink and Stat are separate syscalls against the same magic symlink; a fd-close-and-reuse in between yields a torn read across Name/Deleted vs Type/Device/Size/Node), and noting it's accepted as low-severity, read-only, and metadata-only rather than fixed.

PR #564, opened 2026-07-22
fix(allowedpaths): resolve ".." against the walked path, not lexically, after a symlink

### LINKED ISSUE #561: allowedpaths: .. path segments are collapsed lexically, not resolved through symlinks
### Describe the bug

`toAbs` (allowedpaths/sandbox.go) resolves relative paths via `filepath.Join(cwd, path)`,
which lexically collapses `..` segments before any filesystem/symlink resolution happens.
This diverges from POSIX/kernel path walking, which resolves `..` against the *resolved*
location after following any preceding symlink component.

### Reproduce

`ln -s /tmp/outside /tmp/sandbox/link`
`cat link/../file`   # reads /tmp/sandbox/file, not /tmp/outside/file as GNU cat/the kernel would

Confirmed the same behavior on `cat` and `rm`; likely affects every builtin that resolves
a path through the sandbox (ls, echo redirection, etc.), since they all share `toAbs`/`resolve()`.

For `rm` this doesn't create a cross-sandbox deletion risk in the case tested (it targets
the in-sandbox file, not the one outside), but the general divergence from real path
resolution semantics could matter for other commands or attack scenarios involving
symlinked intermediate components followed by `..`.

Fixing this properly means reworking path resolution in `allowedpaths` to walk `..`
against the filesystem component-by-component (resolving symlinks as it goes) rather
than lexically collapsing the whole path up front — a cross-cutting change affecting
every builtin, not a single-command fix.

Originally raised via Codex review on #556: https://github.com/DataDog/rshell/pull/556#discussion_r3623057327

julesmcrt:
## Summary
- Fixes #561: `toAbs` used `filepath.Join`, which collapses `..` lexically before any symlink component is examined. For a path like `link/../file` where `link` resolves outside the sandbox, this discarded `link` entirely and produced `<cwd>/file` instead of the kernel-correct resolution through `link`'s actual target.
- Read-only operations (`Access`, `Open` reads, `Stat`, `Lstat`, `Readlink`, `ReadDir*`) now use a new `resolveAbsPath` that walks the path one component at a time, following a symlink component only when a later `..` in the path could actually pop through it. Every other symlink component is left untouched so existing escape detection (`os.Root`'s native "path escapes from parent", the cross-root `resolveFollowingSymlinks` fallback) and its error messages are unaffected.
- Write operations (`Open` with write flags, `Truncate`, `TruncateToZeroIfAtLeast`) keep the plain lexical `toAbs`: `writeopen`'s `O_NOFOLLOW` walk needs literal symlink component names intact to detect and reject symlink write targets — eagerly resolving them here would defeat that check.
- Added `TestDotDotAfterSymlinkFollowsKernelSemantics` and `TestDotDotAfterSymlinkEscapeDenied` covering the exact scenario from the issue (both the correct-resolution and escape-denied cases).

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (full suite, including `allowedpaths` and `analysis` symbol-allowlist checks)
- [x] `gofmt -l .` / `go vet ./...` clean
- [ ] `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash` — not run locally (Docker unavailable in this environment); no scenario YAML files were changed by this fix, only Go code in `allowedpaths/`, so this should be safe but worth confirming in CI.

### DIFF
--- a/allowedpaths/sandbox.go
+++ b/allowedpaths/sandbox.go
@@ -438,14 +438,17 @@ func containsDotDot(components []string) bool {
 	return false
 }
 
-// splitComponents splits path into non-empty components using the native
-// separator, without collapsing "." or ".." — callers walk those explicitly
-// so that ".." is applied against a resolved location rather than the raw
-// string.
+// splitComponents splits path into non-empty components, without collapsing
+// "." or ".." — callers walk those explicitly so that ".." is applied
+// against a resolved location rather than the raw string. path is
+// normalized with filepath.FromSlash first so that "/"-separated input
+// (which shell scripts may use even on Windows) splits the same way as
+// native-separator input.
 func splitComponents(path string) []string {
 	if path == "" {
 		return nil
 	}
+	path = filepath.FromSlash(path)
 	parts := strings.Split(path, string(filepath.Separator))
 	out := make([]string, 0, len(parts))
 	for _, p := range parts {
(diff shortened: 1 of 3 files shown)


julesmcrt:
@codex review

julesmcrt:
@codex review

PR #566, opened 2026-07-23
feat(builtins): add pmap builtin for per-process memory-map diagnostics

Hitsuji-M:
## Summary
- Adds a `pmap` builtin reporting per-process virtual memory mappings (address range, size, permission mode, mapping label; `-x` adds per-mapping RSS/Dirty).
- Linux backend reads `/proc/<pid>/{comm,maps,smaps}` directly, bypassing `AllowedPaths` (same documented exception as `ss`/`ip route`/`df`/`free`) since the path is derived only from the numeric PID.
- Windows backend enumerates committed regions via `VirtualQueryEx`, labeling by `MEMORY_BASIC_INFORMATION` type; `-x` is unsupported there.
- macOS has no backend (`pmap: not supported on this platform`) — no cgo-free API exposes per-process memory regions.
- Never reads process argv/environment; header line shows only the comm/executable name.

## Test plan
- [x] `make fmt` / `gofmt -l .` clean
- [x] `go build ./...`
- [x] `go vet ./...`, plus cross-compiled `go vet`/`go test -c` for `GOOS=linux` and `GOOS=windows`
- [x] `go test ./analysis/... ./builtins/...` (allowlists + unit tests)
- [x] `go test ./tests/... -run TestShellScenarios`
- [x] Manual CLI smoke test of `pmap -h`, `help pmap`, `pmap <pid>` on macOS (not-supported path), and rejected-flag paths


builtins/internal/procmaps/procmaps.go
```diff
@@ -0,0 +1,118 @@
...
+// per-mapping Rss/Dirty). Windows enumerates the process's committed
+// virtual memory regions with VirtualQueryEx and labels each region by its
+// MEMORY_BASIC_INFORMATION type ("[image]", "[ mapped ]", "[ anon ]")
+// rather than a resolved file path — per-region file paths require
+// GetMappedFileNameW, which golang.org/x/sys/windows does not wrap, and
+// per-region Rss/Dirty needs a per-page working-set walk that neither
+// package exposes; Read returns ErrExtendedNotSupported for extended mode
+// on Windows rather than reporting fabricated zeros.
+//
+// macOS is not supported: enumerating another process's memory regions
```
matt-dz: Could we support basic macOS before landing this? The public `proc_pidinfo` API exposes `PROC_PIDREGIONINFO`/`PROC_PIDREGIONPATHINFO`, and `SYS_PROC_INFO` is available on Darwin, so region enumeration does not inherently require `task_for_pid` or cgo. This backend can fit behind the existing `readImpl` abstraction; `-x` could remain unsupported until the resident/dirty counter semantics are validated. At minimum, this rationale should be reworded as an implementation gap, but given that the project supports macOS I think basic support is worth including here.

builtins/pmap/pmap_test.go
```diff
@@ -0,0 +1,223 @@
...
+func TestPmapLinuxHappyPath(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Skip("pmap /proc backend is linux-only")
+	}
+	selfPID := os.Getpid()
+	stdout, stderr, code := runScript(t, "pmap "+strconv.Itoa(selfPID))
+	assert.Equal(t, 0, code)
+	assert.Empty(t, stderr)
+	assert.Contains(t, stdout, strconv.Itoa(selfPID)+":")
+	assert.Contains(t, stdout, "total")
```
matt-dz, line 148: Could we add a deterministic fake-`ProcPath` test that asserts the complete basic and extended output, including total Size/RSS/Dirty? These `Contains` checks would still pass if column formatting or the totals were wrong. The same fixture can include a sentinel secret in `cmdline` and assert it is never emitted, locking down the documented argv-privacy guarantee.

builtins/internal/procmaps/procmaps_windows.go
```diff
@@ -0,0 +1,160 @@
...
+		return "", fmt.Errorf("read process name: %w", err)
+	}
+	return filepath.Base(windows.UTF16ToString(buf)), nil
+}
+
+// protectToMode converts a PAGE_* protection constant into pmap's 5-char
+// Mode column (read, write, execute, then two fixed dashes — Windows has
+// no shared/private distinction analogous to Linux's maps 's'/'p' flag at
+// this level of detail).
+func protectToMode(protect uint32) string {
```
matt-dz, line 126: Could we add a `procmaps_windows_test.go` with table tests for `protectToMode` (including modifier bits) and `regionName`? The live Windows test only checks for the PID header and `total`, so incorrect permission or region labels would currently pass.

matt-dz (changes requested):

julesmcrt (commented):
Did a security/correctness pass on top of @matt-dz's review — didn't see overlap with the three points already raised (macOS gap, weak test assertions, missing Windows unit tests), so adding these as supplementary, non-blocking notes.

**Overall:** the `AllowedPaths`-bypass exception is accurately documented and matches what the code does, PID handling is safe against traversal, and there's no argv/environment exposure anywhere in the new code — verified against `docs/RULES.md`'s process-introspection and environment-privacy rules.

A few minor, non-blocking observations:

1. **`builtins/internal/procmaps/procmaps_windows.go` (`readImpl`, the `VirtualQueryEx` loop):** `PROCESS_VM_READ` is requested alongside `PROCESS_QUERY_INFORMATION`. It's genuinely needed by `GetModuleFileNameEx`, and `ReadProcessMemory` is never called, so no process memory *content* is exposed — just flagging that the requested handle rights are broader than the enumeration itself strictly needs, worth a one-line comment noting why.

2. **Same loop:** `MaxMappings` bounds only the *committed* regions appended to the result slice — there's no separate cap on total loop iterations over non-committed regions. In practice this is bounded by the real address-space layout (each iteration strictly advances `addr`) and the 30s exec timeout, so it's not a practical DoS, but it's asymmetric with the Linux backend's `maxLineBuf` defensive ceiling. Might be worth a brief comment explaining why no separate iteration cap is needed here.

3. **Symlink trust model:** unlike the `journalctl` exception (which explicitly rejects symlinked machine directories/journal files), `pmap` doesn't verify `comm`/`maps`/`smaps` aren't symlinks before reading. This matches `ps`'s existing trust model for `ProcPath` being a genuine procfs, so I don't think it needs code changes — just flagging it so it's an explicit, reviewed decision rather than an implicit one, maybe worth a one-line note in the package doc comment alongside the existing sandbox-bypass rationale.

None of these block merging as far as I can tell — nice, defensively-coded addition overall (bounded reads, overflow guards throughout, provably-terminating Windows loop, accurate docs).

builtins/internal/procmaps/procmaps.go
Hitsuji-M: fixed in [1caaedc](https://github.com/DataDog/rshell/pull/566/commits/1caaedccc2783b54f219dde7afa0548930a734fa), same for the other comments

matt-dz (approved):
nicely done 🙌

PR #574, opened 2026-08-04
fix(break/continue): unwind all enclosing loops on out-of-range count

julesmcrt:
## Summary
- `break`/`continue` with an out-of-range count (0 or negative, e.g. `break -1`) only unwound the innermost enclosing loop instead of terminating all nested loops like real bash.
- Fixed by returning a large `BreakN` sentinel for the invalid-count path so the existing clamp-at-outermost logic in `interp/runner_exec.go` fully unwinds the loop stack.

## Test plan
- [x] `make fmt`
- [x] `go test ./builtins/... ./tests/...`
- [x] Manually verified `break -1`/`continue -1`/`break 0`/`continue 0` against real bash
- [x] Added 5 new scenario tests under `tests/scenarios/shell/for_clause/break_cont/`

### DIFF
--- a/builtins/internal/loopctl/loopctl.go
+++ b/builtins/internal/loopctl/loopctl.go
@@ -11,9 +11,10 @@ import (
 	"github.com/DataDog/rshell/builtins"
 )
 
-// unwindAllLoops is used as a break/continue depth that is guaranteed to
-// exceed any realistic loop nesting depth, so the clamp-at-outermost logic
-// in interp/runner_exec.go fully unwinds every enclosing loop. It is a plain
+// unwindAllLoops is used as a break/continue depth that comfortably exceeds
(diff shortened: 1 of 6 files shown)


julesmcrt:
@codex review

builtins/internal/loopctl/loopctl.go
```diff
@@ -11,6 +11,14 @@ import (
 	"github.com/DataDog/rshell/builtins"
 )
 
+// unwindAllLoops is used as a break/continue depth that is guaranteed to
```
astuyve: Guaranteed eh?

astuyve (approved):

builtins/internal/loopctl/loopctl.go
julesmcrt: Strong wording indeed, lowering it a bit

PR #575, opened 2026-08-04
fix(loopctl): match bash exit code 2 for invalid break/continue argument

julesmcrt:
## Summary
- `break`/`continue`'s "numeric argument required" fatal error returned exit code 128; real bash returns 2 for this error.
- Fixed the exit code and corrected the scenario tests that baked in the wrong (128) value, including one in a different directory (`tests/scenarios/shell/while_clause/continue_overflow.yaml`) that a prior pass missed.
- Added new overflow-argument scenarios to pin the behavior.

## Test plan
- [x] `make fmt`
- [x] `go test ./builtins/... ./tests/...`
- [x] Manually verified against real bash 5.3 (`break abc`, `break <huge number>`, `continue <huge number>`) — all exit 2

### DIFF
--- a/builtins/internal/loopctl/loopctl.go
+++ b/builtins/internal/loopctl/loopctl.go
@@ -25,7 +25,7 @@ func LoopControl(callCtx *builtins.CallContext, name string, args []string) buil
 		parsed, err := strconv.Atoi(args[0])
 		if err != nil {
 			callCtx.Errf("%s: %s: numeric argument required\n", name, args[0])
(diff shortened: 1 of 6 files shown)


julesmcrt:
@codex review

astuyve (approved):

julesmcrt:
Don't want to upgrade the entire CI image to use BASH 5.3
Don't want to remove the test against BASH
So leaving the error code `128` that matches BASH 5.2

PR #577, opened 2026-08-04
fix(free): match procps-ng's field-width-driven -h promotion boundary

julesmcrt:
## Summary
- `free -h`'s `humanBytes()`/`formatScaled()` promoted to the next IEC unit only at the true 1024 boundary, unlike real procps-ng `free -h`, which promotes based on a 5-character field-width budget — so a value like 1010Ki rendered as "1010Ki" instead of "1.0Mi".
- Reworked the promotion logic to try the decimal form, then a truncated-integer form, escalating to the next unit whenever neither fits within 5 characters, matching procps-ng's `scale_size()`.

## Test plan
- [x] `make fmt`
- [x] `go test ./builtins/free/...`
- [x] Added boundary test cases (999Ki, 1000Ki→1.0Mi, 1010Ki→1.0Mi, 1023Ki→1.0Mi) to `TestHumanBytes`

### DIFF
--- a/builtins/free/free.go
+++ b/builtins/free/free.go
@@ -287,36 +287,74 @@ func humanBytes(v uint64) string {
 	return formatScaled(val, suffixIdx, suffixes)
 }
 
-// formatScaled renders val (already scaled into [1, 1024) at suffixIdx)
-// with GNU free's one-decimal-below-10 convention.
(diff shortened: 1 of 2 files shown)


julesmcrt:
@codex review

astuyve (approved):

PR #581, opened 2026-08-04
test(interp): add end-to-end coverage for journalctl --rotate

julesmcrt:
## Summary
- Test-coverage gap: `journalctl --vacuum-time` had a real end-to-end integration test (`TestJournalctlVacuumEndToEnd`), but `--rotate` only had a builtin-level test with a fully stubbed `AuthorizeSystemd`/`JournalRotator`, never exercising the real backend.
- Added `TestJournalctlRotateEndToEnd`, which runs `journalctl --rotate` through a real `Runner` in remediation mode against a real Unix-socket fixture speaking the actual `io.systemd.Journal` Varlink framing — exercising real machine-ID validation, real pinned-socket dial, and real request/response framing.
- Added `TestJournalctlRotateDeniedWithoutRemediationDespiteGrant`, which configures a full grant plus the real fixture target with remediation mode off, sharpening the existing `rotate_read_only.yaml` scenario (which can't distinguish "gate denied" from "backend never configured").
- Sanity-checked by disabling the remediation gate — the new denial test failed as expected, then reverted.

No production code changes — the gate was already correctly enforced.

## Test plan
- [x] `make fmt`
- [x] `go test ./interp/...` (run inside a Linux container via colima, since rotate's real implementation is Linux-only)
- [x] Sanity check: disabled the gate locally, confirmed new denial test fails, reverted

### DIFF
--- a/interp/journal_rotate_integration_test.go
+++ b/interp/journal_rotate_integration_test.go
@@ -1,0 +1,192 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-present Datadog, Inc.
+
+//go:build linux
+
+package interp
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


julesmcrt:
@codex review

matt-dz (approved):

PR #586, opened 2026-08-04
docs(systemd): document slice journal read amplification

julesmcrt:
A `read` grant on a `.slice` unit makes `journalctl -u <slice>` match `_SYSTEMD_SLICE`, which journald stamps on every entry produced by every unit in that slice. An operator granting `system.slice:read` to see the slice unit's own messages has in fact granted journal read over every service in that slice — undocumented, and in tension with the "exact grants, never whole-host enumeration" invariant the systemd design advertises.

The behaviour is faithful to upstream `journalctl -u <slice>` and is **not** changed here: removing the match would make slice queries misleadingly empty. This is a docs change plus one pinning test.

- `docs/RULES.md`: new paragraph in the trusted-systemd-target section, next to the existing indirect-effects prose, contrasting the bounded non-slice matches with the unpaired `_SYSTEMD_SLICE` match.
- `AGENTS.md`: matching entry in the "Security Design Decisions" list.
- `SHELL_FEATURES.md`: one clause on the `AllowedSystemServices` policy line, since that is where operators read about grants.
- `internal/systemd/journal_query_file_test.go`: `TestJournalFileQuerySliceGrantReturnsEveryMemberUnit` builds a fixture with two units in one slice, one in a nested child slice, and one elsewhere, and asserts the slice grant returns both members and nothing else.

Note: the match is exact, not transitive — journald records a process's *immediate* slice, so a `system.slice` grant does not reach units in `system-<child>.slice`. The docs and the test both state it that way.


julesmcrt:
@codex review

louis-cqrl (approved):

matt-dz (approved):

PR #588, opened 2026-08-04
fix(allowedpaths): normalize ENXIO on write opens to "not a regular file"

julesmcrt:
A FIFO (or device node) write target made `truncate` / `logrotate` emit a raw, non-portable errno that also changed depending on whether a reader happened to be attached:

```
# no reader
truncate: "/tmp/rsh/fifo": openat fifo: device not configured   # macOS ENXIO; Linux says "no such device or address"
# reader attached
truncate: "/tmp/rsh/fifo": truncate /tmp/rsh/fifo: not a regular file
```

That is non-deterministic output and it escapes `PortableErr` — both violations of `docs/RULES.md`.

## Change

- New `writeopen.ErrNotRegularFile` sentinel; `writePathError` maps `unix.ENXIO` to it (ENXIO on a write open always means a non-regular target, which rshell never accepts here).
- `PortableErrMsg` renders it as `not a regular file`.
- `Sandbox.Truncate` / `Sandbox.TruncateToZeroIfAtLeast` rewrap it as the same `*os.PathError` their post-open `fstat` guard already produces.

Only `ENXIO` is remapped — `errors.Is(err, os.ErrNotExist)` is untouched, so `truncate -c`'s silent-skip path still works. Windows build is unaffected (`unix.ENXIO` is only referenced from `writeopen_unix.go`).

Both branches now print `truncate: "pipe": not a regular file`.

## Tests

The existing FIFO pentest test asserted only exit code 1 and a path substring, and used a read-only sandbox path — so it was passing on `permission denied` and never reached the FIFO logic. It now grants `:rw` and asserts the exact message, plus reader-attached variants and the equivalent pair for `logrotate` (no FIFO coverage existed there). The reader end is opened with `O_RDONLY|O_NONBLOCK` and closed via `t.Cleanup`, so there is no goroutine to hang or flake.

### DIFF
--- a/allowedpaths/internal/writeopen/errors.go
+++ b/allowedpaths/internal/writeopen/errors.go
@@ -12,6 +12,18 @@ import "errors"
 // resolution and open.
 var ErrSymlinkWriteTarget = errors.New("symlinks are not supported as write targets")
 
+// ErrNotRegularFile reports that a write target is not a regular file.
+//
+// It is returned both by the post-open fstat guard and by the open itself
+// when the kernel refuses the open with ENXIO. ENXIO is not exclusively
+// "FIFO opened O_WRONLY|O_NONBLOCK with no reader attached" — some device
+// nodes report it too — but every ENXIO case here describes a non-regular
+// target, which is the only thing rshell's write paths accept. Normalizing
+// it keeps the message identical regardless of whether a reader happened to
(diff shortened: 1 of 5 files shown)


julesmcrt:
@codex review

matt-dz (approved):

PR #589, opened 2026-08-04
fix(interp): create redirect files with 0666 & ~umask

julesmcrt:
`interp/runner_redir_remediation.go` hardcoded `redirectFilePerm = 0644` for files created by `>`, `>|`, `>>`, `&>`, and `&>>`. That is the *result* of `0666 & ~umask` for umask 022 only. bash passes `0666` to `open(2)` and lets the kernel apply the umask — and `Sandbox.Truncate` in this same codebase already does exactly that, so the two write paths in one runner disagreed.

At umask 002:

```
redir.txt  -rw-r--r--   rshell redirect  (wrong)
trunc.txt  -rw-rw-r--   rshell truncate  (correct)
bashredir  -rw-rw-r--   bash             (correct)
```

## Changes

- `redirectFilePerm` is now `0666`; both call sites already share the constant, so this covers all five operators.
- New umask tests for every redirect operator asserting `0644` at umask 022 and `0664` at umask 002. Only the 002 case catches the bug — verified the new test fails against the old constant and passes after the fix. `syscall.Umask` is process-global, so the test is non-parallel and restores the prior value with `defer`.

## Windows

Inert, confirmed rather than assumed: `writeopen_windows.go` calls `os.Root.OpenFile`, which reaches `syscall.Open`, whose only use of `perm` is `if perm&S_IWRITE == 0 { attrs = FILE_ATTRIBUTE_READONLY }`. Both 0644 and 0666 set 0200, so the attribute is unchanged. A Windows counterpart test asserts the files are created and not read-only.

## Testing

- `go test ./interp/... ./allowedpaths/... ./builtins/tests/truncate/...` — pass.
- `go test ./tests/` — pass.
- `RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash` — pass (25.9s).

### DIFF
--- a/interp/runner_redir_remediation.go
+++ b/interp/runner_redir_remediation.go
@@ -16,8 +16,13 @@ import (
 )
 
 // redirectFilePerm is the permission bits used when creating redirect target
-// files. Matches the default umask-applied mode produced by bash.
-const redirectFilePerm os.FileMode = 0644
+// files. 0666 lets the process umask determine the final mode (open(2) applies
+// mode & ~umask), matching bash's >FILE behaviour and Sandbox.Truncate.
+// Hardcoding 0644 here would instead bake in the result for umask 022 only,
+// producing a non-group-writable file on hosts with a 002 umask.
+// On Windows there is no umask; Go maps the mode to the read-only attribute
+// (perm&0200), which both 0644 and 0666 set, so this value is inert there.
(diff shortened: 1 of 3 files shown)


julesmcrt:
@codex review

matt-dz (approved):

PR #590, opened 2026-08-04
test(writeopen): add direct tests for the remediation write primitives

julesmcrt:
`allowedpaths/internal/writeopen` had **0.0%** test coverage despite being the package every mutating operation bottoms out in — `truncate`, `logrotate`, `rm`, and every `>`/`>>`/`&>` redirect all reach `writeopen.OpenFile` or `writeopen.Unlink`. It was exercised only transitively through `allowedpaths`, so the individual rejection branches were unpinned.

Test-only change; no production code touched.

**Coverage: 0.0% -> 94.6%** (93.5% when running as root, where the read-only-parent `unlinkat` case is skipped).

What's covered:
- `OpenFile` — nil `rootFile`, root-directory targets, `..` traversal, symlinked intermediate and final components, the dangling-symlink create attack (explicitly asserts the referent is not created), `perm.Perm()` masking, nested happy path.
- `Unlink` — directory rejection pinned at the primitive level so Linux and macOS agree despite macOS `unlinkat(2)` succeeding on empty directories; `..` traversal; symlinked intermediates; live / dangling / self-referential symlink leaves (link removed, referent untouched); the `requiresDir` / `statFlags = 0` dereference branch.
- `HasTrailingDirSyntax` — table test plus the platform-conditional backslash branch, selected by `runtime.GOOS` since `filepath.Separator` is a compile-time constant.
- Two fd-count regression guards over the `closeDir` bookkeeping.

Separate `writeopen_windows_test.go` for the Windows implementation (different internals, and symlink creation needs a privilege CI runners lack — reasoning is in the file header). Errors are asserted with `errors.Is` against sentinels and errnos, never message strings.

No bugs found. Four defensive branches are unreachable by construction (`filepath.Clean` collapses any `..` that follows a real component, so the `closeDir` cleanup inside the `..` checks can never fire; likewise the `os.NewFile == nil` guard). They are correct as written — the file header documents why they stay uncovered so a reviewer doesn't hunt for a missing test.

Verified: `go test ./allowedpaths/...` passes on darwin and in a `golang:1.25` Linux container; `GOOS=windows go build ./... && GOOS=windows go vet ./allowedpaths/...` clean.

### DIFF
--- a/allowedpaths/internal/writeopen/dirsyntax_test.go
+++ b/allowedpaths/internal/writeopen/dirsyntax_test.go
@@ -1,0 +1,61 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-present Datadog, Inc.
+
+// These are Go tests by necessity: writeopen is an internal package and the
+// YAML scenario framework cannot reach it directly, so AGENTS.md's "prefer
+// scenario tests" rule does not apply here.
+
+package writeopen
+
+import (
+	"path/filepath"
+	"runtime"
+	"testing"
+)
+
+func TestHasTrailingDirSyntax(t *testing.T) {
+	tests := []struct {
+		name    string
+		relPath string
+		want    bool
+	}{
+		{name: "empty", relPath: "", want: false},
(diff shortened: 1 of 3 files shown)


julesmcrt:
@codex review

matt-dz (approved):

PR #591, opened 2026-08-04
test(systemctl): add end-to-end scenario coverage for mutations, grant shapes, and list filters

julesmcrt:
Only ten `systemctl` scenarios existed. Every mutation verb except `start` and every list filter had no end-to-end coverage through the real runner (argument parsing, `AllowedSystemServices` policy, remediation gating, output formatting).

Adds 14 scenarios:

- **Mutation denials** — `stop`, `reload`, `restart`, `enable`, `disable`, mirroring `start_denied`.
- **Grant shapes** — read does not imply mutate; a `start` grant does not imply `read` (for `status` or `list-units`); a grant for `api.service` does not match `api` or `api.socket`.
- **Operand validation** — `list-units` rejects operands; the 32-operand bound (including that it is checked pre-deduplication); the 255-byte name bound; malformed names (`foo.bogus`, `foo`, `.service`, `foo.`, `@foo.service`, `a@b@c.service`, empty, none); duplicates are deduplicated, not rejected.
- **List filters** — `--type` / `--state` / `--all` / `--no-legend`, comma-separated and repeated forms, unsupported and empty values, and the 32-value bound.

Also adds an `allowed_system_services` field to the scenario harness (`tests/scenarios_test.go`) so grant-shape denials can be expressed declaratively. It maps directly to `interp.AllowedSystemServices` and defaults to no grants, so every existing scenario is unchanged. Test-only — no production code touched.

Paths that need a live systemd manager (a successful `start`, real `status` output, a populated `list-units`, and `--state` filtering, which runs on manager replies) are not expressible as scenarios and stay in `builtins/systemctl/systemctl_test.go` with its mock backend.

All new scenarios set `skip_assert_against_bash: true`.

### DIFF
--- a/tests/scenarios/cmd/systemctl/basic/duplicate_operands.yaml
+++ b/tests/scenarios/cmd/systemctl/basic/duplicate_operands.yaml
@@ -1,0 +1,15 @@
+# Repeated operands are deduplicated rather than rejected: validation raises no
+# duplicate error and the run reaches the normal authorization denial.
+skip_assert_against_bash: true
+description: systemctl accepts and deduplicates repeated unit operands
+input:
+  mode: remediation
+  script: |+
+    systemctl status api.service api.service api.service
+    systemctl restart api.service api.service
+expect:
+  stdout: |+
+  stderr: |+
+    systemctl: system service "api.service" is not allowed for action "read"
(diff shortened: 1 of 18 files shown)


julesmcrt:
@codex review

matt-dz (approved):

PR #596, opened 2026-08-04
feat(rm): bound file removals with a cumulative per-run budget

julesmcrt:
## Problem

`rm`'s `MaxRemoveFiles = 10` cap is per-invocation, so it bounds only a single mistaken glob. `for f in *; do rm "$f"; done` and `find … | xargs -n1 rm` each issue one single-operand invocation per file and delete an unbounded number of files without ever tripping it. Nothing in the remediation surface bounded a run in aggregate.

## Change

Audit finding 02, Option A (the recommendation): add a run-wide cumulative budget and keep the per-invocation cap.

- `builtins.MaxFileRemovalsPerRun = 100`, charged inside the sandbox `Remove` capability so it covers `rm` however it is dispatched (direct, loop, subshell, `xargs`, `find -exec`).
- Backed by `Runner.fileRemovalCount`, an `atomic.Int64` created per `Run()` and shared by pointer with every subshell and pipeline stage — the same shape as the existing `globReadDirCount`.
- Only successful removals are charged; a failed unlink refunds its slot, so a script hammering a missing path cannot burn an operator's allowance. Reserve-then-refund (rather than check-then-increment) keeps concurrent pipeline stages from overshooting the cap.
- `rm` reports an exhausted budget with a distinct message naming the limit and stops processing remaining operands instead of repeating the same error.
- Package doc, `rm --help`, and `SHELL_FEATURES.md` updated.

## Open for maintainer sign-off

- **The number (100).** Ten is clearly too low for a run-wide budget; 100 is large enough for a realistic log/temp cleanup while still bounding an unattended run to damage an operator can reason about. It is a guess, not a measured figure.
- **Fixed constant vs. `RunnerOption`.** Implemented as a fixed constant: one number every deployment shares is harder to misconfigure, and no caller has yet shown a bulk-cleanup need. Happy to convert it to a `RunnerOption` if operators need it.

## Deliberately not done

- The budget covers deletions only. `truncate` / `logrotate` mutations still have no count limit; extending the same mechanism to them is a separate decision.
- No `AGENTS.md` security-design entry — that list documents `AllowedPaths` bypasses, and this is not one.

## Tests

- `tests/scenarios/cmd/rm/basic/run_budget_loop.yaml` and `run_budget_xargs.yaml` — the two bypass idioms, both bounded at 100 (`skip_assert_against_bash: true`; real bash has no such limit).
- `interp/rm_run_budget_test.go` — runner-state plumbing YAML cannot express: counter installed by `Run()`, shared with ordinary and background subshells, reset between `Run()` calls, and not charged for failed removals.
- Existing per-invocation cap scenarios (`cap_boundary`, `too_many_files`) unchanged and passing; `hardening/help` extended for the new help text.
- `go test ./...` green. `TMPDIR=$HOME/.rshell-tmp RSHELL_BASH_TEST=1 go test ./tests/ -run TestShellScenariosAgainstBash` green (14s).


julesmcrt:
@codex review

matt-dz (approved):

PR #598, opened 2026-08-04
fix(rm): clear NoOptDefVal sentinel before printing help

julesmcrt:
## Summary
- `rm --help` was leaking a raw NUL byte into its output for `--help`/`--verbose`, because `RegisterNoArgBool`'s NUL sentinel (`NoOptDefVal`) wasn't cleared before `fs.PrintDefaults()` — pflag only suppresses inline rendering of `NoOptDefVal` for bool flags when it equals `"true"`.
- Applies the same fix pattern already used in `stat.go`/`df.go` and just landed in `logrotate.go` (PR #593 review comment): clear the sentinel on matching flags via `fs.VisitAll` right before printing defaults, then restore it via `defer`.

## Test plan
- [x] Added `TestRmHelpContainsNoNulByte`, verified it fails against the pre-fix code (via `git stash`) and passes with the fix
- [x] `make fmt` (clean)
- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./builtins/tests/rm/...`
- [x] `cmd/rm/*` scenario tests

### DIFF
--- a/builtins/rm/rm.go
+++ b/builtins/rm/rm.go
@@ -101,6 +101,23 @@ func registerFlags(fs *builtins.FlagSet) builtins.HandlerFunc {
 			callCtx.Out("Remove each FILE. Directories are never removed.\n")
 			callCtx.Out("Symlinks are removed without following them.\n")
 			callCtx.Outf("At most %d files may be removed per invocation.\n\n", MaxRemoveFiles)
+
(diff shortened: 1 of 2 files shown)


julesmcrt:
@codex review

matt-dz (approved):

PR #603, opened 2026-08-21
Tag the run span with the raw command and effective interp options

julesmcrt:
## Summary
- Add `interp.Script(text string) RunnerOption` to attach the raw, unparsed script/command string to a `Runner`, and tag it on the top-level `run` telemetry span as `rshell.run.command`.
- Tag the effective `RunnerOption` configuration on the same span as `rshell.run.options.*`: `mode`, `max_execution_time`, `proc_path`, `host_prefix`, `allowed_paths` (as `path:ro`/`path:rw`), `allow_all_commands`, `allowed_commands`, `allowed_system_services`, and `systemd_target_configured`.
- Wire `cmd/rshell/main.go` to pass the raw script text through `interp.Script` so the CLI's telemetry includes it.
- Add `rshell.run.invoked_via_cli`, tagged unconditionally on the same span, so operators can tell runs launched through the `cmd/rshell` binary apart from `interp` embedded as a library.
- Before the raw command text is tagged, regex-based scrubbing (`interp/command_scrub.go`) redacts common secret shapes (URL-embedded credentials, `Authorization` headers, `key=value`/`--flag value` pairs whose name looks sensitive, AWS access key IDs, JWT-shaped tokens, etc.) and replaces them with a fixed `********` placeholder. This is best-effort defense-in-depth, not a guarantee. Operators whose commands are too sensitive for any residual risk should use `DisableDetailedTelemetry` instead.

Example trace showing the tagged span: https://app.datadoghq.com/u/d4735e3a/tbS-nf7-ShJ

## Test plan
- [x] `make fmt`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./interp/... ./cmd/...` (added `TestRunSpanCommandAndOptions`, `TestRunSpanInvokedViaCLI` in `interp/tracing_test.go`; scrubbing covered by `TestScrubCommandText`/`TestScrubCommandTextDoesNotLeakInputValues` in `interp/command_scrub_test.go`)

### DIFF
--- a/interp/api.go
+++ b/interp/api.go
@@ -110,6 +110,13 @@ type runnerConfig struct {
 	// [Script]; empty when the caller does not supply one.
 	commandText string
 
+	// disableDetailedTelemetry, when true, suppresses the rshell.run.command
+	// and rshell.run.options.* tags on the top-level "run" telemetry span.
+	// The rshell.version, rshell.run.exit_code, and other outcome tags set in
+	// [Runner.Run] are unaffected. Set via [DisableDetailedTelemetry]; false
+	// (tags emitted) by default.
+	disableDetailedTelemetry bool
+
 	// remediationMode enables remediation-only capabilities, including file-target
 	// output redirections within AllowedPaths and the restricted systemctl builtin.
 	remediationMode bool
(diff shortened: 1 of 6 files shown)


interp/api.go
```diff
@@ -102,6 +103,27 @@ type runnerConfig struct {
 	// Defaults to "/proc" when empty.
 	procPath string
 
+	// commandText holds the raw, unparsed shell script or command string this
```
matt-dz: nit: can we make these doc comments more concise? they're filled with a bit too much detail and will have to be edited whenever behavior changes. it's also one of my least favorite vibe code artifacts 😅

interp/api.go
```diff
@@ -700,6 +727,53 @@ func (r *Runner) Run(ctx context.Context, node syntax.Node) (retErr error) {
 	return nil
 }
 
+// setRunOptionTags records the effective [RunnerOption] configuration of r on
+// the top-level "run" span. All fields read here are set once during [New]
+// and never mutated afterwards, so this is safe to call without additional
+// synchronization from within Run.
+func (r *Runner) setRunOptionTags(span *telemetry.Span) {
```
matt-dz, line 724: nit: same here

interp/api.go
```diff
@@ -881,6 +955,46 @@ func ProcPath(path string) RunnerOption {
 	}
 }
 
+// Script attaches the raw, unparsed shell script or command string that this
```
matt-dz: nit: a lot of these functions as well. a bit odd to have comments take more lines than the actual function itself

matt-dz (approved):
looks good. let's clean up some of the long essays claude wrote for comments

PR #606, opened 2026-09-02
feat: add sha256sum builtin

matt-dz:
## Summary

Adds a native, sandboxed `sha256sum` builtin for computing and verifying SHA-256 file digests.

This enables agents to record a config file’s digest during planning, then verify that the file is unchanged before executing a human-approved follow-up command:

```sh
printf '%s\n' '<digest>  /path/to/config' |
  sha256sum --check --status &&
  sed '42s/X/Y/' /path/to/config
```

If the file changed, `sha256sum` exits non-zero and the command after `&&` does not run.

## Features
- Hash files or standard input.
- Verify GNU-style checksum manifests with -c / --check.
- Support --quiet, --status, and --help.
- Accept tagged and untagged manifests, escaped filenames, CRLF input, and uppercase digests.
- Enforce AllowedPaths and regular-file-only access.
- Bound manifest size, line length, entry count, operands, and verification output.
- Honor context cancellation, including blocked stdin reads on Windows.

This provides compare-before-edit protection, but is not an atomic compare-and-write primitive; a small TOCTOU window remains between verification and the subsequent command.

## Validation
- Full Go test suite with the race detector.
- GNU/bash scenario comparison.
- Linux, macOS, and Windows CI.
- Compliance and static-analysis tests.
- Fuzz and adversarial security coverage.


matt-dz:
@codex review

builtins/sha256sum/sha256sum.go
```diff
@@ -0,0 +1,657 @@
...
+			continue
+		}
+		if totals.entries == validBefore {
+			callCtx.Errf("sha256sum: %s: no properly formatted checksum lines found\n", builtins.SafeOperand(sourceLabel(source)))
+			failed = true
+		}
+	}
+
+	if !opts.status && totals.entries > 0 {
+		writeCheckWarnings(callCtx, totals)
```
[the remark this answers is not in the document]
matt-dz: Fixed in d17b5836 by tracking warning counters per manifest and emitting each manifest’s warnings before advancing. Regression coverage: tests/scenarios/cmd/sha256sum/check/multiple_manifests.yaml. Verified with focused tests, the GNU/bash scenario comparison, and go test ./....

matt-dz:
@codex review

builtins/sha256sum/sha256sum.go
```diff
@@ -0,0 +1,708 @@
...
+
+	r, err := callCtx.OpenRegularFile(ctx, path)
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+	return fn(r)
+}
+
+func withCancellableReader(ctx context.Context, r io.Reader, fn func(io.Reader) error) error {
```
julesmcrt, line 465: P3 - Duplicates existing cancellation-fallback logic

`withCancellableReader`'s goroutine-per-blocked-read fallback (`cancellableReader`) reimplements the same deadline-probe + goroutine-fallback pattern already in `builtins/read/read.go` for cancelling blocking stdin reads. Consider extracting a shared helper in a future pass so the two implementations don't drift.

builtins/sha256sum/sha256sum.go
```diff
@@ -0,0 +1,708 @@
...
+		if line == "" {
+			continue
+		}
+
+		entry, ok := parseChecksumLine(line)
+		if !ok {
+			totals.malformed++
+			continue
+		}
+		if manifestIsStdin && entry.name == "-" {
```
julesmcrt, line 319: P3 - as a manifest entry name reads real stdin

When a checksum manifest is itself a file (not stdin) and lists an entry named -, `digestPath` reads from the process's actual stdin rather than a literal file named -. This mirrors GNU behavior closely enough to not be a bash-compat bug, but a short comment noting the side channel (a file-based manifest can still consume the script's stdin) would help future readers.

analysis/symbols_builtins.go
```diff
@@ -612,6 +612,29 @@ var builtinPerCommandSymbols = map[string][]string{
...
+		"errors.New",                      // 🟢 creates bounded read/validation errors; pure function, no I/O.
+		"fmt.Errorf",                      // 🟢 bounded error formatting; pure function, no I/O.
+		"io.EOF",                          // 🟢 sentinel error value; pure constant.
+		"io.Reader",                       // 🟢 interface type; no side effects.
+		"os.ErrDeadlineExceeded",          // 🟢 sentinel returned when the stdin read deadline fires; used only to synchronize with context cancellation.
+		"strings.Builder",                 // 🟢 bounded filename escaping and unescaping; pure in-memory buffer.
+		"strings.HasPrefix",               // 🟢 detects checksum format prefixes; pure function, no I/O.
+		"strings.NewReader",               // 🟢 wraps empty in-memory stdin; no external I/O.
+		"strings.ToLower",                 // 🟢 normalizes hexadecimal digest text; pure function, no I/O.
+		"time.Time",                       // 🟢 deadline value used to make blocking stdin reads cancellable; pure data type.
```
julesmcrt, line 632: P3 - Redundant symbol allowlist entries

`time.Time` and `time.Unix` are already present in the global `builtinAllowedSymbols` slice (used by other commands). Listing them again here is harmless but redundant.

julesmcrt (approved):
LGTM, nits can be skipped

PR #612, opened 2026-09-10
fix(privileged-helper): normalize overlapping path modes

astuyve:
### What does this PR do?
Consider an EP with the following config:
```
/:ro
/:rw
```

Both refer to /. Because Landlock permissions are additive, their actual combined permission is read-write. So now we normalize them to be:
```
  /:rw
```
  - RW wins regardless of ordering:

    ["/:ro", "/:rw"] → ["/:rw"]
    ["/:rw", "/:ro"] → ["/:rw"]

  - It only merges identical normalized paths. It does not merge parent/child paths:

    ["/tmp:rw", "/tmp/readonly:ro"]

   allowing Landlock to reject that unrepresentable configuration.

  - It runs after local intersection so RW cannot bypass local policy. An RW entry can only survive if both the signed backend policy and local helper policy permit RW.
  - Invalid paths are preserved so the sandbox rejects them with its existing fail-closed validation instead of silently dropping them.

### Checklist

- [ ] Tests added/updated
- [ ] Documentation updated (if applicable)

### DIFF
--- a/privilegedhelper/verify.go
+++ b/privilegedhelper/verify.go
@@ -314,8 +314,11 @@ func intersectPaths(requested, configured []string) []string {
 	return result
 }
 
-// normalizeEffectivePaths merges duplicate paths, preferring read-write.
-// Parent and child paths remain distinct for Landlock validation.
+// normalizeEffectivePaths merges duplicate paths, preferring read-write:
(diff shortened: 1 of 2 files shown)


matt-dz (approved):
comments are a bit long but logic is sound

astuyve:
yeah let me tighten those up, the bot got a little excited when I asked it to clearly explain why this was necessary

PR #615, opened 2026-09-10
feat: narrow privileged execution with an unsigned Agent-supplied policy layer

julesmcrt:
### What does this PR do?

Adds a third, middle intersection layer to the privileged helper's authorization model: an unsigned, Agent-supplied `AgentPolicy` carried on `ExecuteRequest`, representing the Private Action Runner's `datadog.yaml`-configured `restricted_shell.allowed_*` operator settings. Today those settings only narrow the non-privileged rshell path; the privileged path only had backend allowlists intersected with an optional root-owned `policy.json`.

New model:
```
backend (signed task) allowlists ∩ AgentPolicy (optional, from datadog.yaml via the Agent) ∩ optional local policy.json
```

- `privilegedhelper/protocol.go`: new `AgentPolicy` type (`AllowedCommands`, `AllowedPaths`, `AllowedSystemServices`, `ElevatableCommands`) attached to `ExecuteRequest.AgentPolicy *AgentPolicy`. Fields deliberately have no `omitempty` tag so nil (JSON `null`) and non-nil-empty (JSON `[]`/`{}`) survive the wire round-trip distinctly.
- `privilegedhelper/verify.go`: `Verify()` now checks each of the four `AgentPolicy` fields for nil **independently** before intersecting that axis. A nil field defers entirely to signed ∩ policy.json (no narrowing); a non-nil field (even empty) narrows/denies via the existing `intersectCommands`/`intersectPaths`/`intersectSystemServices`/`intersectExact` helpers. `authorizationContext` diagnostics gained an `Agent` field alongside `Signed`/`Local`/`Effective`.
- `docs/PRIVILEGED_HELPER.md`: documents the new layer and its per-field nil-vs-empty convention.

### Motivation

`policy.json` is root-owned and optional, so most deployments never author one — meaning privileged execution today is effectively bounded only by the signed backend task. Operators who already restrict the non-privileged path via `datadog.yaml` had no way to apply the same restriction to privileged/elevated commands. This closes that gap without weakening the trust boundary: `AgentPolicy` is unsigned and Agent-supplied, so it can only **narrow**, never grant — even a compromised/buggy Agent process can at most fail to narrow something, never expand beyond what the signed task + policy.json already allow.

### Security note

Per-field nil-vs-empty semantics matter here: if a whole-struct nil/non-nil check were used instead, configuring just one operator setting (e.g. `allowed_commands`) would have silently denied every other axis (e.g. system services) on the privileged path only, diverging from non-privileged behavior in a way that would be easy to miss and hard to debug. This PR checks each field independently and includes a wire round-trip test guarding the `omitempty` pitfall (Go's `encoding/json` serializes nil and empty slices/maps identically when `omitempty` is set, which would have destroyed the distinction over the actual Unix socket even though in-process tests wouldn't catch it).

### Testing

- Unit tests in `verify_test.go`: no-AgentPolicy regression parity, partial-field AgentPolicy only narrows configured axes, explicit-empty-field deny-all, true 3-way intersection across all three layers, and an over-permissive AgentPolicy cannot widen beyond the signed task.
- Wire round-trip test in `protocol_test.go` (`TestAgentPolicyWireRoundTripPreservesNilVsEmpty`) guarding the `omitempty` pitfall.
- Full repository test suite green, including `-race`.

### Companion PR

DataDog/datadog-agent#(pending) wires the Private Action Runner's existing `restricted_shell.allowed_*` config plus a new `restricted_shell.privileged.elevatable_commands` setting into this new `AgentPolicy` field. That PR currently depends on this one merging and a new rshell version being published before it builds against a real (non-replace-directive) dependency.

### Checklist

- [x] Tests added/updated
- [x] Documentation updated

> **AI assistance disclosure:** This PR was developed with assistance from AI coding agents (Pi subagents). The author reviewed and validated the resulting code and documentation.

### DIFF
--- a/docs/PRIVILEGED_HELPER.md
+++ b/docs/PRIVILEGED_HELPER.md
@@ -35,6 +35,44 @@ They authorize `journalctl` reads in either mode and `systemctl` or journal
 cleanup only in remediation mode. `elevatableCommands` remains a separate
 requirement when an operation needs effective UID 0.
 
+## Authorization layers
+
+The effective policy for a privileged request is the intersection of up to
+three independent layers, applied in this order: the signed backend task,
+the unsigned Agent-supplied policy carried on the request (see
+"Agent-supplied policy" below), and the optional local root-owned
+`policy.json` (see "Optional local policy" below). Any layer that is absent
+imposes no narrowing; a present layer only ever narrows what the preceding
+layers already allow.
+
+## Agent-supplied policy
+
+Every `ExecuteRequest` may carry an unsigned `agentPolicy` object alongside
+the signed envelope. This is how the Datadog Agent's Private Action Runner
+forwards its own `private_action_runner.restricted_shell.allowed_commands`,
+`allowed_paths`, and `allowed_system_services` `datadog.yaml` settings (plus
+an elevatable-commands equivalent) to the privileged path — the same
+operator settings that already narrow the non-privileged rshell path.
+
+Because `agentPolicy` is unsigned and supplied by a process the helper does
+not fully trust, it can only narrow the signed backend policy and any local
+`policy.json` — it can never grant a permission beyond what those already
+allow. A request with no `agentPolicy` at all behaves exactly as if this
+field never existed: the effective policy is signed task ∩ optional
+`policy.json`.
+
+When `agentPolicy` is present, each of its four fields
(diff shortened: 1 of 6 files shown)


astuyve (approved):
Image

had a super goofy UI bug but I think it's just github, gofmt would ban this

